PR #7, opened 2021-04-26
Add Pytorch MNIST Example

hanlint:
Adds a working pytorch MNIST example, following the new spec. 
```
I0425 17:19:27.115946 139984905910080 submission_runner.py:220] Final mnist_pytorch score: 9.956104
```

Since `pytorch` uses classes to store state, whereas `jax` is more functional, the key change to the `spec.py` needed is to extend `ParameterTree` type to include `torch.nn.Module`. We will need to resolve a better naming for this type, as it is no longer a `Tree`. 

The parameter names should also change in some of the spec functions. For example, in `update_params`, the name `current_params` is no longer accurate... since this is a pytorch module, we can call its functions such as:
* `current_params(input)` - run forward pass
* `current_params.eval()` - switch to eval mode

One proposal (unimplemented) to fix this is to rename `ParameterTree` -> `Model`, and `current_params` -> `current_model`.

A few TODOs to resolve with JAX (cc: @znado )

- [x] match model architecture in the JAX implementation
- [ ] when tensor/model loading to device occurs
- [x] create shared `workload` class between JAX/PyTorch for common properties
- [x] resolve RNG between JAX/PyTorch
- [x] standardize JAX/Pytorch on a single `eval_model` function


hanlint:
OK great, I'll look over your changes. A few thoughts on the above:

**Naming**
`current_param_container` is clunky, but I can't think of an alternative. 

** tensor/model loading **
The principled question is whether we allow submissions to alter the device loading logic. If we keep this in `update_params`, users are free to do work on the tensors on host if needed, before loading to cuda (albeit with some performance impacts). If we enforce that `update_params` must operate on cuda-loaded tensors, then the logical location is in the`preprocess` fixed function. `init_optimizer_state` would be an unexpected place for this. 

Given that the `model.to('cuda')` is already in a fixed function, I would recommend we keep the tensor loading also in a fixed function.

I'm OK with using the JAX rng. I'm not familiar with how it works --  Is it safe to just use half of the PRNGKey, as I did in the code, or should I reconstruct the 64-bit?

There's one more item upon review: can we add a `data-dir` to the argparse to handle the datasets location?

hanlint:
@znado I merged your branch, and added some additional fixes to get all working (see commits above). I think this is ready to merge, pending reviews from @mikerabbat. Recommend we stage the naming / device loading as a separate PR pending discussions in the working group.

znado:
That sounds good to me! We can keep the current names and do a follow-up that just does renaming.

**Device placement**
I think we may have two device placement decisions: when to place model ops/vars and when to place optimizer ops/vars. I think it seems reasonable that we always have the model be on device, but allow the submitters to do whatever they want with device placement concerning the optimizer variables; it doesn't seem unlikely that someone could want to distribute optimizer variables differently than the model. So then I think your current `model.to('cuda')` seems to make sense, I'll update the jax accordingly in a follow-up.

**RNGs**
Taking the first 32 bits of the Jax seed should be fine I believe, with a caveat. If we initialize with just 32 bits then if you grab `rng[0]` _before doing any folding/splitting_, the PRNGKey implementation will pad the 32 bits with zeros so `rng[0]` will be `0`; I'll add a `rng, _ = jax.random.split(rng, 2)` to ensure the two halves are properly mixed in a follow-up change to be safe.

A data_dir flag sounds great, thanks for adding!

setup.py
```diff
@@ -0,0 +1,49 @@
+"""MLCommons Algorithmic Efficiency.
+
+For a Jax GPU install:
+pip install -e .[jax-gpu] -f https://storage.googleapis.com/jax-releases/jax_releases.html'
```
mikerabbat, line 4: Minor: Missing single quote at the beginning of the URL

mikerabbat (approved):
Looks great to me overall!

I believe the Pytorch model encapsulates `current_params` and `model_state`. Would it make sense to have a single container for both that gets unpacked in the Jax version?

Minor comment, fwiw: in submission_runner.py, it's not currently possible to run with `tuning_ruleset == 'self'` because `hyperparameters` and `rng` aren't defined within the scope of the else block starting at line 267.

hanlint:
OK great, I think this PR is ready to merge then, and we will have some follow-on work to standardize:
* device loading
* naming convention
* rng safety

@mikerabbat : Yeah, might be something worth talking about. In our `spec.py`, the `ModelState` and Params` are almost always passed together, the only exception being `data_selection`:
https://github.com/mlcommons/algorithmic-efficiency/blob/e4406e574f4372a5025a9cc4e853d675fc4334de/spec.py#L274-L281

So we might have an opportunity for simplification. However, a clear separation between the "parameter" container, and "other auxiliary info" might be more natural to JAX users (?) than a tuple.

mikerabbat:
Good point @hanlint about `model_state` and `current_params` almost always being passed together. If the motivation for passing `current_params` to `data_selection`, along with `workload`, is to allow the data selection function to potentially call `model_fn` on data to evaluate the loss or something else, then we probably ought to pass `model_state` there too. I'll create a separate issue for this.

PR #11, opened 2021-05-26
[WIP] JAX ResNet50 Imagenet

danielsnider:
Adds a JAX ResNet Imagenet workload. The code is based on the [Flax ImageNet Example v0.3.3](https://github.com/google/flax/tree/v0.3.3/examples/imagenet).

#### Roadmap for this PR to be ready to merge:
- [ ] Need appropriate hyperparameters set in `workloads/imagenet/imagenet_jax/config.py` or `workloads/imagenet/imagenet_jax/tuning_search_space.json`.
- [ ] Need appropriate accuracy target for `has_reached_goal` and need to verify convergence. The [ResNet paper](https://www.cv-foundation.org/openaccess/content_iccv_2015/papers/He_Delving_Deep_into_ICCV_2015_paper.pdf) and the [flax example](https://github.com/google/flax/tree/v0.3.3/examples/imagenet) report between 73% to 77% top-1 accuracy
- [ ] Need help with 2 TODOs in `workload.py:model_fn()`.
- [ ] Standardize JAX/Pytorch implementations.

#### Also keep in mind:
- Tested with 1, 2, and 4 RTX 2080 Ti GPUs. 80% utilization to nvidia-smi.
- A new dependency [ml_collections](https://github.com/google/ml_collections) is used in config.py.
- On my system I had to run with NCCL_SHM_DISABLE=1. Not sure why.
- The `max_allowed_runtime_sec` and eval time (in workload.py) are set low for testing purposes.
- Also for testing purposes, Imagenette is used with ResNet1. This can be configured in config.py.

#### Example run:
```
    --workload=imagenet_jax \
    --submission_path=workloads/imagenet/imagenet_jax/submission.py \
    --tuning_ruleset=external \
    --tuning_search_space=workloads/imagenet/imagenet_jax/tuning_search_space.json \
    --num_tuning_trials=1
```

#### Example output:
```
I0526 18:51:32.759134 139935589717824 submission_runner.py:253] --- Tuning run 1/1 ---
/home/dans/.local/lib/python3.6/site-packages/jax/lib/xla_bridge.py:317: UserWarning: jax.host_count has been renamed to jax.process_count. This alias will eventually be removed; please update y
our code.
  "jax.host_count has been renamed to jax.process_count. This alias "
I0526 18:51:33.311938 139935589717824 dataset_builder.py:877] No config specified, defaulting to first: imagenette/full-size-v2
I0526 18:51:33.313256 139935589717824 dataset_info.py:354] Load dataset info from /home/dans/tensorflow_datasets/imagenette/full-size-v2/0.1.0
/home/dans/.local/lib/python3.6/site-packages/jax/lib/xla_bridge.py:304: UserWarning: jax.host_id has been renamed to jax.process_index. This alias will eventually be removed; please update your
 code.
  "jax.host_id has been renamed to jax.process_index. This alias "
I0526 18:51:33.315797 139935589717824 dataset_builder.py:570] Constructing tf.data.Dataset for split train[0:9469], from /home/dans/tensorflow_datasets/imagenette/full-size-v2/0.1.0
I0526 18:51:43.914318 139935589717824 workload.py:281] train epoch: 0, step: 1, loss: 2.7532, accuracy: 11.72
I0526 18:51:43.975103 139935589717824 dataset_builder.py:877] No config specified, defaulting to first: imagenette/full-size-v2
I0526 18:51:43.976839 139935589717824 dataset_info.py:354] Load dataset info from /home/dans/tensorflow_datasets/imagenette/full-size-v2/0.1.0
I0526 18:51:43.978225 139935589717824 dataset_builder.py:570] Constructing tf.data.Dataset for split validation[0:3925], from /home/dans/tensorflow_datasets/imagenette/full-size-v2/0.1.0
I0526 18:51:45.537803 139935589717824 workload.py:313] eval epoch: 0, loss: 2.7962, accuracy: 8.69
I0526 18:51:45.538616 139935589717824 submission_runner.py:208] 9.07s   1       0.08203125
I0526 18:52:14.039853 139935589717824 workload.py:281] train epoch: 6, step: 447, loss: 1.8603, accuracy: 35.89
I0526 18:52:15.695460 139935589717824 workload.py:313] eval epoch: 6, loss: 1.4383, accuracy: 51.68
I0526 18:52:15.707166 139935589717824 submission_runner.py:208] 39.10s  447     0.5011836886405945
I0526 18:52:44.041692 139935589717824 workload.py:281] train epoch: 12, step: 907, loss: 1.3763, accuracy: 54.85
I0526 18:52:45.312337 139935589717824 workload.py:313] eval epoch: 12, loss: 1.2163, accuracy: 60.44
I0526 18:52:45.327490 139935589717824 submission_runner.py:208] 69.12s  907     0.5849294066429138
I0526 18:53:03.022012 139935589717824 submission_runner.py:267] Tuning trial 1/1
I0526 18:53:03.022305 139935589717824 submission_runner.py:268] Hyperparameters: Hyperparamters(learning_rate=0.00018805814126031094, one_minus_beta_1=0.9261041856291229, epsilon=1e-08)
I0526 18:53:03.024841 139935589717824 submission_runner.py:269] Metrics: {'eval_results': [(1, 0.08203125), (447, 0.5011836886405945), (907, 0.5849294066429138)], 'global_step': 1197}
I0526 18:53:03.024910 139935589717824 submission_runner.py:270] Timing: 60.02908158302307
I0526 18:53:03.024973 139935589717824 submission_runner.py:271] ====================
I0526 18:53:03.025090 139935589717824 submission_runner.py:302] Final imagenet_jax score: 60.029082
```

I welcome and appreciate all comments!


danielsnider:
I've requested to sign the CLA.

workloads/imagenet/imagenet_jax/submission.py
```diff
@@ -0,0 +1,177 @@
...
+import ml_collections
+
+import spec
+from . import config as config_lib
+
+config = config_lib.get_config()
+
+
+
+# flax.struct.dataclass enables instances of this class to be passed into jax
```
znado, line 20: I think we can delete TrainState (here and in workload.py), and pass around these three attributes separately wherever this is passed; it looks like all the places we use this wrapper class (which I believe is only in update_params), we already have access to the optimizer state and the global step

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,314 @@
...
+      model_state=model_state)
+    return params, state
+
+  def model_fn(
+      self,
+      params: spec.ParameterTree,
+      augmented_and_preprocessed_input_batch: spec.Tensor,
+      model_state: spec.ModelAuxillaryState,
+      mode: spec.ForwardPassMode,
+      # rng: spec.RandomState, # TODO: Question— I'm not sure how to pass rng in
```
znado, line 222: that's a very good question! random state in a multidevice setting can be confusing (even moreso in a multihost, but we don't have to worry about that). in Jax we can just pass the same rng to each device, and then fold in the device index (example [here](https://github.com/google/init2winit/blob/master/init2winit/trainer.py#L151)) in order to get a per-device rng. we could also split the rng into `num_devices` rng's, and pass one to each device (but IIRC this is slower due to weird Jax issues)

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,314 @@
...
+  def _build_dataset(self,
+      data_rng: jax.random.PRNGKey,
+      split: str,
+      data_dir: str,
+      batch_size):
+    if config.batch_size % jax.device_count() > 0:
+      raise ValueError('Batch size must be divisible by the number of devices')
+    local_batch_size = config.batch_size // jax.host_count()
+
+    if config.half_precision:
```
znado, line 58: we'll always run in float32, so you can delete this logic

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,314 @@
...
+from flax import optim
+from flax.training import common_utils
+
+import spec
+from . import input_pipeline
+from . import models
+from . import submission
+from . import config as config_lib
+
+config = config_lib.get_config()
```
znado, line 27: we can delete this (see comments below)

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,314 @@
...
+
+  def has_reached_goal(self, eval_result: float) -> bool:
+    return eval_result > 0.6
+
+  def _build_dataset(self,
+      data_rng: jax.random.PRNGKey,
+      split: str,
+      data_dir: str,
+      batch_size):
+    if config.batch_size % jax.device_count() > 0:
```
znado, line 54: we can just use the batch_size passed into this function

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,314 @@
...
+      else:
+        input_dtype = tf.float16
+    else:
+      input_dtype = tf.float32
+
+    ds_builder = tfds.builder(config.dataset)
+    ds = input_pipeline.create_input_iter(
+      ds_builder,
+      local_batch_size,
+      config.image_size,
```
znado, line 70: we can hardcode image_size to 224 in the input pipeline, because we'll always be running with the same data preprocessing

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,314 @@
...
+      input_dtype,
+      train=split=='train',
+      cache=config.cache)
```
znado, line 73: I would hardcode this to True instead, I believe it should always improve performance

znado:
I think this is a really good start! Lets chat more about this PR this week (I emailed you for times). For a few of your original questions:

- I would use the hparams for Nesterov momentum in table 1 of [our paper](https://arxiv.org/abs/2102.06356) (ignoring \gamma_0). You can find a Jax implementation of the LR schedule [here](https://github.com/google/init2winit/blob/master/init2winit/schedules.py#L154).
- I would set an accuracy target of >75%, we should discuss at the WG meeting the exact number, the MLPerf training competition uses 75.9% and maybe we should go with that? I would hold off on this because it will make debugging very hard.
- I think we can remove the need for ml_collections.ConfigDict, see PR comments
- not sure about `NCCL_SHM_DISABLE`, we can look into that!

danielsnider:
Thanks for the review, Zack. 💯 Your comments were super helpful! I have resolved the easier issues. I'll take a look at the rest tomorrow morning.  When I'm feeling ready to chat about the remaining improvements, I'll let you know. :-)

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,314 @@
...
+        submission.train_step,
+        model.apply,
+        learning_rate_fn=learning_rate_fn,
+        loss_fn=self.loss_fn,
+        loss_type=self.loss_type,
+        model_fn=self.model_fn,
+        compute_metrics=self.compute_metrics),
+      axis_name='batch')
+
+    self.p_eval_step = jax.pmap(
```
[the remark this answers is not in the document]
danielsnider: In the process of addressing this comment I got a strange harmless error shown below. It just prints at the end of the run and everything else seems to work. But I can't find anything on Google and I've tried a lot to fix it. It was introduced by this commit [f8d6374](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/f8d6374028e296405094a8b5e5e4cfc1ef12e283). I tried [this commit](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/333d142be5f6101a76b159c511c1b10fffd49909) and [that commit](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/04994c4fea92307c213c5429c1226d849bdbcf23) but I couldn't get the error to go away.
```
I0614 18:00:38.480022 140028703885120 submission_runner.py:270] ====================
I0614 18:00:38.480246 140028703885120 submission_runner.py:301] Final imagenet_jax score: 20.000402
Exception ignored in: <bound method IteratorResourceDeleter.__del__ of <tensorflow.python.data.ops.iterator_ops.IteratorResourceDeleter object at 0x7f59b055fb48>>
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/data/ops/iterator_ops.py", line 533, in __del__
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/eager/context.py", line 2031, in eager_mode
  File "/usr/lib/python3.6/contextlib.py", line 159, in helper
TypeError: 'NoneType' object is not callable
```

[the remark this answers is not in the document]
danielsnider: I'm going to move onto other improvements and come back to this at the end.

workloads/imagenet/imagenet_jax/input_pipeline.py
```diff
@@ -0,0 +1,254 @@
...
+
+  return jax.tree_map(_prepare, xs)
+
+
+def create_input_iter(dataset_builder, batch_size, image_size, dtype, train, cache):
+  ds = create_split(
+      dataset_builder, batch_size, image_size=image_size, dtype=dtype,
+      train=train, cache=cache)
+  it = map(prepare_tf_data, ds)
+  it = jax_utils.prefetch_to_device(it, 2)
```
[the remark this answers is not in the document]
danielsnider: I will investigate the performance effect of `prefetch_to_device` but first I am tracking down which recent commit caused a slow down from 80% GPU utilization down to 50%.

danielsnider:
Hi @znado,

All of your suggestions were great! Thank you very much.

Here's what I have left to do. A little bit of guidence from you would be appreciated :-).

- I will investigate the performance effect of [prefetch_to_device](https://github.com/mlcommons/algorithmic-efficiency/pull/11#discussion_r646861601) but first I am trying to understand why [this recent commit, "delete TrainState and pass around these three attributes separately"](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/0e5d659033d0c4da5197c124f5a26f76b80024a0) caused a slow down from 80% GPU utilization down to 50%. I tried reverting the spirit of the commit by [reintroducing TrainState](https://github.com/danielsnider/algorithmic-efficiency/commit/2c8cac62030767cf22277a139a863d58df5affda) ontop of the latest code, but I didn't see a performance boost. Let me know if you have any thoughts, otherwise I will try whatever I can think of.
- I will use the hparams for Nesterov momentum from table 1 of [your paper](https://arxiv.org/abs/2102.06356) (ignoring \gamma_0). I will replace my [learning_rate_fn](https://github.com/mlcommons/algorithmic-efficiency/pull/11/files#diff-7fff18c0f3f15c1ad38a70e36f97076a9b4980e6b8f59f1df7bece4b6b7491c6R26) with the [LR schedule implementation](https://github.com/google/init2winit/blob/master/init2winit/schedules.py#L154) that you gave me. That's correct right? Then I want test how fast/close I can get to the 75.9% accuracy target.
- Lastly, I am stuck trying to find a solution to [this harmless error](https://github.com/mlcommons/algorithmic-efficiency/pull/11#discussion_r647086925).

I'm hoping you can give me a suggestion about the harmless error, and that I can fix the performance issue. Also, in a follow-up PR, we can make a shared parent class with common ImageNet code.

Many thanks, talk to you soon.

workloads/imagenet/imagenet_jax/workload.py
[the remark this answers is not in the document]
znado: That is a...bizarre error! I've never seen it unfortunately, and I was about to suggest that you do those two exact changes you tried with those two commits. I would say lets ignore it for now and see if it's still there when we merge this PR, and if so I can ask the TF/Jax support groups for help diagnosing it.

znado:
> * I will investigate the performance effect of [prefetch_to_device](https://github.com/mlcommons/algorithmic-efficiency/pull/11#discussion_r646861601) but first I am trying to understand why [this recent commit, "delete TrainState and pass around these three attributes separately"](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/0e5d659033d0c4da5197c124f5a26f76b80024a0) caused a slow down from 80% GPU utilization down to 50%. I tried reverting the spirit of the commit by [reintroducing TrainState](https://github.com/danielsnider/algorithmic-efficiency/commit/2c8cac62030767cf22277a139a863d58df5affda) ontop of the latest code, but I didn't see a performance boost. Let me know if you have any thoughts, otherwise I will try whatever I can think of.

That's strange that the change would cause that! My first suspicion was relating to `pmap`ing, but this commit doesn't actually touch that. We must have moved something in the process, maybe regarding when we replicate state, maybe at [this line](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/0e5d659033d0c4da5197c124f5a26f76b80024a0#diff-7fff18c0f3f15c1ad38a70e36f97076a9b4980e6b8f59f1df7bece4b6b7491c6L138) or [this line](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/0e5d659033d0c4da5197c124f5a26f76b80024a0#diff-2774bdbee47f6dc3f4dcef8f3d6f448c703215226e131c62994bec18abe93656R186)? To me it seems most natural to replicate the model state inside `init_model_fn` like you have it though, so if this causes the slowdown we should investigate. As a simple (hacky) check you could try not replicating anything until you first hit [this line](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/0e5d659033d0c4da5197c124f5a26f76b80024a0#diff-7fff18c0f3f15c1ad38a70e36f97076a9b4980e6b8f59f1df7bece4b6b7491c6L138), and seeing if that recovers the utilization? If it isn't that, we should print out the shapes or pytree structures of the various state objects being passed around before/after the commit to make sure we aren't accidentally double replicating something.

Also, to confirm, if you check out that older commit, you can still repro getting 80% GPU utilization?

> * I will use the hparams for Nesterov momentum from table 1 of [your paper](https://arxiv.org/abs/2102.06356) (ignoring \gamma_0). I will replace my [learning_rate_fn](https://github.com/mlcommons/algorithmic-efficiency/pull/11/files#diff-7fff18c0f3f15c1ad38a70e36f97076a9b4980e6b8f59f1df7bece4b6b7491c6R26) with the [LR schedule implementation](https://github.com/google/init2winit/blob/master/init2winit/schedules.py#L154) that you gave me. That's correct right? Then I want test how fast/close I can get to the 75.9% accuracy target.

Yeah that should be a solid baseline for ImageNet for now, we will need to come up with generic submissions once we have more workloads implemented but for the sake of testing the code this should work.

> * Lastly, I am stuck trying to find a solution to [this harmless error](https://github.com/mlcommons/algorithmic-efficiency/pull/11#discussion_r647086925).

That's a weird one! Let's hope it goes away, and if not I can try to ask around and see what the deal is.

danielsnider:
#### GPU utilization 

For the GPU utilization slowing down to 50%. I was investigating the wrong commit because I am intermittently seeing different utilizations. The most repeatable slowdown to 50% is the [commit "switch from flax optimizer to optim"](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/b3695c53eb1d0db39104e23a10a5812d97b11ae3). I think changing the optimizer is more significant so perhaps this makes sense? 

Can you help me understand why I might be seeing different utilizations intermittently?
How trustworthy are the numbers from nvidia-smi? If the GPU temp goes up (from 35C to around 52C), does that trigger rate-limiting and artificially boost the utilization %? Is it possible for XLA to chooses suboptimal strategies before it finds a good one? Does the XLA compilation cache persist between executions? 

#### Learning Rate
I took a look at [your paper](https://arxiv.org/abs/2102.06356) table:
image

But I couldn't figure out how map the Nesterov column of your table map to the hparams required by the [lr implementation](https://github.com/google/init2winit/blob/master/init2winit/schedules.py#L154) that you gave me, which requires:

    'warmup_steps'
    'base_lr'
    'start_lr'
    'end_lr'
    'power'
    'warmup_power
    'decay_end'
    
Could you please point me in the right direction?

Many thanks! 
Daniel

znado:
> The most repeatable slowdown to 50% is the commit "switch from flax optimizer to optim".

Aha, I think that commit accidentally changed to using Adam instead of Nesterov, which keeps 2 copies of model parameters (one for each of first and second moment estimates). Can you instead try using [`optax.sgd`](https://github.com/deepmind/optax/blob/master/optax/_src/alias.py#L509), passing `nesterov=True` (and then you can also remove the `optax.chain`).

> How trustworthy are the numbers from nvidia-smi? If the GPU temp goes up (from 35C to around 52C), does that trigger rate-limiting and artificially boost the utilization %?

This I honestly don't know about, but I would assume that temperature doesn't affect utilization unless the GPU got dangerously close to its physical operating limits.

> Is it possible for XLA to chooses suboptimal strategies before it finds a good one?

My understanding is that the XLA compiler will, for given hardware/driver/software versions, give the same compilation for the same program.

> Does the XLA compilation cache persist between executions?

No, it is a totally new cache each time you run.

> I couldn't figure out how map the Nesterov column of your table map to the hparams required by the lr implementation that you gave me, which requires:

```
'base_lr': 7.05,
'warmup_power': 2.0,
'warmup_steps': 706.5,
'end_lr': 0.000006,
'decay_end': 2512,
'power': 2.0,
'start_lr': 0.0,
```

this is assuming you're training at a batch size of 32,768, which is *way* larger than we're training for, and for 2512 steps, which is likely much fewer than we're training for. Nonetheless, my guess is the general shape of this schedule could still be a strong baseline. I would set `decay_end` to something similar to the number of steps you expect it to take to reach the goal, and scale `warmup_steps` proportionally.

danielsnider:
Hi @znado,

Here are some updates. 

**GPU Usage** 
Switching from Adam to SGD gave an increase from 50% to 60% GPU utilization. Switching from ResNet1 to ResNet50 gives a further increase from 60% to 90%. Note that I don't always see these numbers. For determinism, I tried the XLA flags but I didn't see any obvious changes and I didn't spend a lot of time on it.

**Convergence**
Using the hyperparameters that you provided in your last message unfortunately didn't give me good convergence. After two days training ResNet50 on ImageNet with 4x 2080Ti GPUs, I saw a leveling off at 8% train accuracy. Just FYI, ImageNet has 1,281,167 training examples, and our batch size is 128, which works out to 10,008 steps per epoch. I tried 512 batch size but I run out of GPU memory. Right now I am trying a batch size 256, and I increased `decay_end` and `warmup_steps` by 10x from what you gave me (so 25,000 and 7,000 respectively), and I am syncing batch parameters across devices 10 times per epoch (instead of just once). I'll let you know how it's going after a few hours of training.

It might be worth voice chatting about hyperparameter search because I don't have much experience here and I would greatly benefit from a little advice. Some questions that come to mind are: Is it better to finalize our hyperparameters on our actual competition hardware? Is there a hyperparameter range and search technique that you recommend I try? Or is there another published paper with a "good" hyperparameter set that I should try?

Many thanks!
Daniel

danielsnider:
Update: After 2.5 hours of training, my run with 256 batch size is still under 1% eval accuracy 👎.

znado:
Got it. The learning rate (`base_lr`) may be way too high, what were you using? If you're using the `7.05` from the previous comment, that's for a batch size of 32,768, so if we're only running with a batch size of 256, we may want to try tuning the LR in a neighborhood around a 64x smaller learning rate (linearly scaling down the LR by a factor of 32768/256=64).

As for `eval_period_time_sec`, we can set it to eval every ~4 epochs, so I would find whatever the steps/second is and calculate roughly how many seconds pass every 4 epochs. For `max_allowed_eval_time_sec`, it would be nice to aim for 64 epochs til convergence (which is how fast we could converge in), but we should set the max time to be more relaxed, so I would calculate roughly how many seconds 76 epochs is and start there.

danielsnider:
@znado, Thanks for the tips! After doing the math, I should eval every 100 minutes (4 epochs). And convergence should not take more than 31 hours (76 epochs). I will give it a try! I'm sorry I won't have results until next week because I'm having a datacentre hardware issue today and because of our Canada Day long weekend. Hopefully my speed of development progress picks up significantly up next week! Talk to you soon.

danielsnider:
Hi @znado, with this set of hyperparameters and batch size 256, unfortunately ImageNet accuracy doesn't surpass 25% 👎. This is with 4x 2080 Ti GPUs using ResNet50.
```python
hparams = {
    'base_lr': 7.05/64,
    'warmup_power': 2.0,
    'warmup_steps': 706.5*100,
    'end_lr': 0.000006,
    'decay_end': 2512*100,
    'power': 2.0,
    'start_lr': 0.0,
  }
  ```

Perhaps I should try a hyperparameter range and search technique that you recommend? This is my [commit](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/c2de3c64d484b34e651860bdef38af6713072bff) if you want to see exactly what I'm doing currently.

**Comment by znado** [comment]
Ok, maybe we should abandon that hyperparameter setting/LR schedule for now, it likely requires heavier retuning at smaller batch sizes, which I can do later once we know the workload works; I was hopeful we would at least get decent (70%+) accuracies by just using LR/64 without further tuning, but that's OK.

I would try taking the flax imagenet example [hyperparameters](https://github.com/google/flax/blob/master/examples/imagenet/configs/default.py) and [LR schedule](https://github.com/google/flax/blob/master/examples/imagenet/train.py#L95) and use those. Also, for the sake of submitting this PR, for now lets relax the convergence targets to whatever we can achieve with those Flax settings, and then in a follow-up PR lets focus on making the baseline(s) more competitive.

**Comment by danielsnider** [comment]
Hi @znado ,

After 20 hours training on 4x 2080 Ti's at 90% GPU utilization, I reached 52% accuracy and it's still increasing. This is by far the best ImageNet convergence that I've seen. I have to stop it now to do other GPU work today, but I will train for longer this weekend.

After re-reading the flax example I saw new improvements and I added them to this PR in [these commits](https://github.com/mlcommons/algorithmic-efficiency/pull/11#commits-pushed-3798fad). Please take a look.

These are the only outsanding issues that I can think of, and maybe best addressed in follow-up PRs.
- [ ] Standardize and share code between JAX/Pytorch implementations.
- [ ] On my system I had to run with NCCL_SHM_DISABLE=1. Probably just a quark of my hardware.
- [ ] Tuning hyperparameters for competative baseline.

Thanks!

**Comment by znado** [comment]
> After 20 hours training on 4x 2080 Ti's at 90% GPU utilization, I reached 52% accuracy and it's still increasing. This is by far the best ImageNet convergence that I've seen. I have to stop it now to do other GPU work today, but I will train for longer this weekend.

This seems promising! I'll do a final pass over the code, and I'll probably have some more comments, but lets just go ahead and submit this once that's done.

> * [ ]  Standardize and share code between JAX/Pytorch implementations.

Lets do this in a follow-up PR once we have both versions submitted to avoid premature optimization :)

> * [ ]  On my system I had to run with NCCL_SHM_DISABLE=1. Probably just a quark of my hardware.

I haven't seen this before, but I'd add a commented out line that sets this, with a comment saying on your hardware setup that was necessary. What happens if you don't set this? A GPU OOM?

> * [ ]  Tuning hyperparameters for competative baseline.

This will be an adventure just by itself, because we'll need to make baselines that are competitive across all the workloads, so lets definitely save this for later.

**Comment by danielsnider** [comment]
@znado, sounds great. Thanks!

File: workloads/imagenet/imagenet_jax/workload.py [hunk 10]
```diff
@@ -0,0 +1,290 @@
...
+      params: spec.ParameterTree,
+      model_state: spec.ModelAuxillaryState,
+      rng: spec.RandomState,
+      data_dir: str):
+    """Run a full evaluation of the model."""
+
+    # sync batch statistics across replicas once per epoch
+    model_state = self.sync_batch_stats(model_state)
+
+    epoch_metrics = common_utils.get_metrics(self.epoch_metrics)
```
znado, line 246: for now, lets remove all of `epoch_metrics` (here, in `submission.py`, and in `__init__()`), we will need to add something like this later, but I would prefer to make a more universal solution for all workloads so we can have a unified logging pipeline (we'll need consistent logging code and file format so submitters can audit their submission performance if we run it).

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,290 @@
...
+    if self._eval_ds is None:
+      self._eval_ds = self._build_dataset(
+        data_rng, split='test', batch_size=eval_batch_size, data_dir=data_dir)
+    eval_iter = iter(self._eval_ds)
+    total_accuracy = 0.
+    eval_step = 1
+    start_time = time.time()
+    accumulated_eval_time = 0
+    for batch in eval_iter:
+      metrics = jax.pmap(self.eval_model_fn,
```
znado, line 265: instead of having jax.pmap here, can you add it as a function decorator around `eval_model_fn` (similar to what we did with `pmapped_train_step` in `submission.py`)

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,290 @@
...
+                      model_state,
+                      rng)
+      eval_metrics.append(metrics)
+      total_accuracy += jnp.mean(metrics['accuracy'])
+      eval_step += 1
+      # Check if eval time is over limit
+      current_time = time.time()
+      accumulated_eval_time += current_time - start_time
+      is_time_remaining = (
+        accumulated_eval_time < self.max_allowed_eval_time_sec)
```
znado, line 279: I don't think we need to have a `max_allowed_eval_time_sec`, we can just iterate through the whole eval dataset every time and remove the logic around this (and the definition of the property above)

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,290 @@
...
+      if not is_time_remaining:
+        num_batches = eval_step
+        break
+
+    eval_metrics = common_utils.get_metrics(eval_metrics)
+    summary = jax.tree_map(lambda x: x.mean(), eval_metrics)
+    logging.info('eval loss: %.4f, accuracy: %.2f',
```
znado, line 286: lets remove the logging from inside the workload, so we can do it consistently for all workloads inside the submission runner

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,290 @@
...
+                  summary['loss'], summary['accuracy'] * 100)
+    return float(total_accuracy / num_batches)
```
znado, line 288: can you return the `summary` dict of metrics instead of just the accuracy float?

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,290 @@
...
+      # Check if eval time is over limit
+      current_time = time.time()
+      accumulated_eval_time += current_time - start_time
+      is_time_remaining = (
+        accumulated_eval_time < self.max_allowed_eval_time_sec)
+      if not is_time_remaining:
+        num_batches = eval_step
+        break
+
+    eval_metrics = common_utils.get_metrics(eval_metrics)
```
znado, line 284: do we need `common_utils.get_metrics`? It grabs the metrics from the 0-th device and places them on the CPU ([code](https://github.com/google/flax/blob/main/flax/training/common_utils.py#L51)), which is all we should need given `eval_metrics` should be a dict of strings -> np.array of shape (num_devices,), I don't think we need the `stack_forest`. You can probably just replace this with just:
```
  eval_metrics = jax.tree_map(lambda x: x[0], eval_metrics)
  eval_metrics = jax.device_get(eval_metrics)
```

workloads/imagenet/imagenet_jax/input_pipeline.py
```diff
@@ -0,0 +1,262 @@
+# Copyright 2021 The Flax Authors.
```
znado, line 1: here and in `models.py`, remove this copyright and instead add a comment saying it was forked from the flax repo with a link to the file it was forked from. IANAL but my understanding is this is sufficient under the Apache 2.0 license, because this repo is also Apache 2.0

workloads/imagenet/imagenet_jax/input_pipeline.py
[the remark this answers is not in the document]
znado: cool! lets keep it here for now then and leave a comment saying on your hardware it yielded that increase

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,263 @@
...
+    eval_batch_size = 200
+    num_batches = self.num_eval_examples // eval_batch_size
+    if self._eval_ds is None:
+      self._eval_ds = self._build_dataset(
+        data_rng, split='test', batch_size=eval_batch_size, data_dir=data_dir)
+    eval_iter = iter(self._eval_ds)
+    total_accuracy = 0.
+    for idx in range(num_batches):
+      batch = next(eval_iter)
+      metrics = self.eval_model_fn(params, batch, model_state, rng)
```
znado, line 254: small nit: can we rename this to something like `synced_metrics` to indicate that we already did a `lax.pmean`?

znado:
We chatted offline about how to setup the optimizer state and update function, and came up with a fine solution. Once my two minor comments are closed this is good to merge!

workloads/imagenet/imagenet_jax/submission.py
```diff
@@ -0,0 +1,155 @@
...
+      nesterov=True,
+      momentum=hyperparameters.momentum,
+      learning_rate=learning_rate_fn
+    )
+  return opt_init_fn, opt_update_fn
+
+
+def init_optimizer_state(
+    workload: spec.Workload,
+    model_params: spec.ParameterTree,
```
znado, line 56: I think we renamed ParameterTree to ParameterContainer

workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -0,0 +1,263 @@
...
+      rng: spec.RandomState) -> spec.Tensor:
+    return (selected_raw_input_batch, selected_label_batch)
+
+  def preprocess_for_eval(
+      self,
+      raw_input_batch: spec.Tensor,
+      raw_label_batch: spec.Tensor,
+      train_mean: spec.Tensor,
+      train_stddev: spec.Tensor) -> spec.Tensor:
+    return (raw_input_batch, raw_label_batch)
```
znado, line 146: we can delete these two `preprocess_for_train` and `preprocess_for_eval` functions

danielsnider: These are functions in the `spec.py`. Should keep these functions?

```
$ git grep -n -e preprocess_for_eval -e preprocess_for_train
spec.py:154:  def preprocess_for_train(
spec.py:164:  def preprocess_for_eval(
```

znado: Ah yeah go ahead and delete those too, also could you delete them in the [MNIST example](https://github.com/mlcommons/algorithmic-efficiency/blob/main/workloads/mnist/mnist_jax/workload.py#L81) and edit the [submission runner](https://github.com/mlcommons/algorithmic-efficiency/blob/main/submission_runner.py#L175) to get rid of them too (just use `selected_train_{input,label}_batch` below).

danielsnider: Ok. In [my commit](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/3d5e909c58e77365d1f4d4c20cd027ada3d33ad4) I deleted the preprocess furctions from the imagenet workload, spec, and submission_runner but I didn't delete it for MNIST (jax or pytorch implementations) because the [pytorch MNIST actually does preprocessing](https://github.com/mlcommons/algorithmic-efficiency/blob/main/workloads/mnist/mnist_pytorch/workload.py#L109) and the [common MNIST workload uses it](https://github.com/mlcommons/algorithmic-efficiency/blob/main/workloads/mnist/workload.py#L53). It would be nice to apply that pytorch-only preprocessing to at `build_input_queue()` time, but I did not spend enough time to find a way that worked in pytorch. Maybe @hanlint knows. So in my commit for this, I didn't change MNIST usage of preprocess functions, though, I had to [added one preprocess call](https://github.com/mlcommons/algorithmic-efficiency/pull/11/commits/3d5e909c58e77365d1f4d4c20cd027ada3d33ad4#diff-3cf4126d2419ed538704e070b881b435f2d2c9b0ef833cc2583921833e45e450R50) to make up for the removal from submission_runner.py.

I also renamed `augmented_and_preprocessed_input_batch` to `input_batch` because we aren't mandating a preprocessing function call anymore, so it doesn't make sense to always have that name. The spec was requiring it.

znado:
Can you pull the current copy of the main brain to sync to HEAD? (also there's a `input_batch = input_batch in workloads/mnist/mnist_pytorch/submission.py that we can delete)

danielsnider:
> Can you pull the current copy of the main brain to sync to HEAD? (also there's a `input_batch = input_batch in workloads/mnist/mnist_pytorch/submission.py that we can delete)

Okay, I've synced this branch with main. I ran all three workloads to make sure they are all still working. The two jax workloads need the `--use_jax_rng` option, otherwise they get an error. Is that normal? This is the error:
image

PR #17, opened 2021-06-23
Adding a pure numpy impl of PRNG split/fold so we don't require jax installs.

znado:
Also making a small fix for the Jax MNIST submission where `try_gcs=True` was slowing down our tfds reading.

Tested that can run the PyTorch MNIST example without `jax` installed, and that the Jax MNIST example still runs.

The general strategy is to import a seed manipulation module like:
```
import random_utils as prng
try:
  import jax.random as prng
except (ImportError, ModuleNotFoundError):
  logging.warning(
      'Could not import jax.random for the submission runner, falling back to '
      'numpy random_utils.')
```

And then call `prng.split`, `prng.fold_in`, `prng.PRNGKey`.

I'm open to suggestions if people prefer a different strategy for switching which implementation to use (`FLAGS`, etc.)!

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -64,7 +64,10 @@ flags.DEFINE_string(
 flags.DEFINE_boolean(
     'use_jax_rng',
     False,
-    'Whether to use the Jax or Numpy RNG library.')
+    'Whether to use the Jax or Numpy RNG library. For PyTorch users, this flag '
+    'should be set to false (the default) so that Jax is not required for the '
+    'run; in addition to adding another dependency, Jax can be aggressive and '
+    'reserve all GPU memory, causing OOMs).')
 
 FLAGS = flags.FLAGS
 
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -23,12 +23,6 @@ import time
 
 import halton
 import random_utils as prng
-try:
-  import jax.random as prng
-except (ImportError, ModuleNotFoundError):
-  logging.warning(
-      'Could not import jax.random for the submission runner, falling back to '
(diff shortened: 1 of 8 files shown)


hanlint:
Thanks @znado , I am able to be unblocked by this PR. However, if we must maintain separate implementations, I would recommend switching based on a flag. With the current design, I need to explicitly `pip uninstall jax` to access the numpy rng. Otherwise, the pytorch code will silently use the jax rng and hit CUDA OOM.

Now that we have a functionally equivalent numpy rng though, is it still necessary to maintain a jax-based rng for the managing the seeds?

znado:
> I would recommend switching based on a flag

Added `--use_jax_rng`.

> Now that we have a functionally equivalent numpy rng though, is it still necessary to maintain a jax-based rng for the managing the seeds?

The Flax models assume jax.PRNGKeys for their rng manipulation, which is unfortunate, but given the current setup I don't think it should be much maintenance work to keep the jax rngs around.

submission_runner.py
```diff
@@ -66,8 +60,11 @@
 flags.DEFINE_string(
     'data_dir',
     '~/',
-    'Dataset location'
-)
+    'Dataset location')
+flags.DEFINE_boolean(
+    'use_jax_rng',
+    False,
+    'Whether to use the Jax or Numpy RNG library.')
```
hanlint, line 67: Can you just add a 'For Pytorch, use the Numpy RNG library' to the description here to motivate the differences for users?

znado: done!

PR #22, opened 2021-07-27
Make target value a property

### LINKED ISSUE #18: Make target evaluation metric more accessible
Currently, the target value to achieve in each workload is only defined via the `has_reached_goal` function, i.e.

https://github.com/mlcommons/algorithmic-efficiency/blob/584769427aceeac2835c055fae40f159d4a415d7/workloads/mnist/workload.py#L6-L7

As discussed in the rules draft, submissions should have access to the target value, i.e. to adapt the behavior in the self-tuning ruleset once close to this finish line. I believe we could make this value more accessible by directly providing it as a property of the workload. Perhaps like this:

```python
class Mnist(spec.Workload):

  def has_reached_goal(self, eval_result: float) -> bool:
    return eval_result['accuracy'] > self.target_value

  @property
  def target_value(self):
    return 0.9
```

@znado @hanlint

fsschneider:
Fixes #18.

The specs now define an abstract property called `target_value` for the `workload` class.

https://github.com/fsschneider/algorithmic-efficiency/blob/948c5a02336a51749069c2ecd81b343d82393a11/spec.py#L128-L130

The individual workloads specify this value individually but framework-independent:

https://github.com/fsschneider/algorithmic-efficiency/blob/948c5a02336a51749069c2ecd81b343d82393a11/workloads/mnist/workload.py#L11-L13

### DIFF
--- a/spec.py
+++ b/spec.py
@@ -126,6 +126,10 @@ class Workload(metaclass=abc.ABCMeta):
     """The types of the parameters in the workload model."""
 
   @abc.abstractproperty
+  def target_value(self):
+    """The target value to reach."""
+
+  @abc.abstractproperty
   def loss_type(self):
     """The type of loss function."""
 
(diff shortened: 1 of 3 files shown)


znado (approved):
LGTM!

PR #23, opened 2021-07-27
Add workload properties with number of training/eval examples

### LINKED ISSUE #20: Add a workload property with the number of training examples
It would be helpful to have a workload property with the number of training examples.

This would allow submissions to do epoch-related calculations.

_Moved from comment in rules draft_

fsschneider:
This fixes #20 and allows submissions to do epoch-related calculations.

The properties are defined as part of the specs and will then be specified in each workload (but across frameworks).


znado (approved):
LGTM!

PR #24, opened 2021-07-28
Adding a required framework flag so that the user can specify which framework to use.

znado:
Generalizes the previous `use_jax_rng` flag.


fsschneider:
As a follow-up, I think it would be beneficial to automatically select the correct workload based on the framework. I.e., that the user does not have to specify `--workload=mnist_jax` but simply `--workload=mnist` and the `mnist_jax` is selected if `--framework=jax`.

PR #27, opened 2021-08-10
Automatic Speech Recognition Pytorch Workflow

ckt624:
Training Dataset: Librispeech clean 100 hour.
Eval Dataset: Librispeech test_clean
Model: CNN LSTM model.
Platform: 8 V-100 GPU machine.

Character Error Rate: 0.0995 after 52760 steps or 52825.76 seconds (1.0 step/second).


workloads/librispeech/librispeech_pytorch/README.md
```diff
@@ -0,0 +1,30 @@
+## 1. Overview
+
+Speech recognition accepts raw audio samples and produces a corresponding text
+transcription. We train a CNN LSTM model on LibriSpeech clean 100 hour dataset
+and evaliate it on Librispeech test clean dataset. We get Character Error Rate
```
runame, line 5: Typo: 'evaliate' -> 'evaluate'

workloads/librispeech/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,219 @@
...
+        batch_size=batch_size,
+        shuffle=True,
+        pin_memory=True,
+        collate_fn=train_collate_fn)
+
+    self._valid_loader = torch.utils.data.DataLoader(
+        valid_set,
+        batch_size=batch_size,
+        pin_memory=True,
+        collate_fn=train_collate_fn)
```
runame, line 83: Potentially it could be beneficial to set the argument `num_workers` of the `DataLoader` classes to a positive integer (the default is `0`) -- see [here](https://pytorch.org/docs/stable/data.html#single-and-multi-process-data-loading) for the docs. It's not clear how to choose the optimal value, it might best to increase it by `+1` until the training process is not getting faster anymore. It might also not help at all, but it could be worth trying if it hasn't been considered yet.

runame (approved):
I specifically checked the PyTorch related code and everything looks good to me. I did not try to run the code. The data loading process could potentially be improved, see my comment regarding this.

workloads/librispeech/librispeech_pytorch/models.py
```diff
@@ -0,0 +1,267 @@
...
+    log_probs = x.log_softmax(dim=-1).transpose(0, 1)
+
+    return log_probs, output_lengths
+
+  def eval_loss(self, dataset, loss_fn, device):
+    self.eval()
+
+    total_loss = 0.0
+    total_count = 0.0
+    for (_, features, trns, input_lengths) in dataset:
```
znado, line 252: is `trns` a typical abbreviation in the speech field? would it be possible to make it a more explicit variable name?

znado (approved):
This is fantastic, I'm super excited to see this workload get in!

workloads/librispeech/librispeech_pytorch/workload.py
ckt624: We will use single host rather than multi host training, so we don't have to set num_workers?

fsschneider:
@ckt624 @znado @runame 
Could you each give me a signal when this is ready to merge?

workloads/librispeech/librispeech_pytorch/workload.py
runame: Sorry, I don't understand what you mean with "single host training". Do you mean that the entire dataset can be loaded in memory (because it's small enough), or that only a single CPU core is used (because one [suggestion](https://pytorch-lightning.readthedocs.io/en/latest/guides/speed.html#num-workers) is to use `num_workers=os.cpu_count()`)? In case you mean that we are only using data parallelism on a single node with multiple GPUs, setting `num_workers > 0` might still help if I'm not mistaken (assuming that the whole dataset cannot be loaded in memory and that there is more than one CPU core).

kamalkraj:
@ckt624 @znado 

Where exactly are the workload-specific python libraries version requirements are added? 

For example, this workload requires `librosa`, `pandas`, .. libraries

workloads/librispeech/librispeech_pytorch/workload.py
ckt624: I set num_workers=2 after some trials.

workloads/librispeech/librispeech_pytorch/models.py
```diff
@@ -0,0 +1,257 @@
...
+        groups=self.n_features,
+        padding=0,
+        bias=None)
+
+  def forward(self, x):
+    x = x.transpose(0, 1).transpose(1, 2)
+    x = F.pad(x, pad=self.pad, value=0)
+    x = self.conv(x)
+    x = x.transpose(1, 2).transpose(0, 1).contiguous()
+    return x
```
kamalkraj, line 146: Not used.

ckt624: Removed.

ckt624:
> @ckt624 @znado
> 
> Where exactly are the workload-specific python libraries version requirements are added?
> 
> For example, this workload requires `librosa`, `pandas`, .. libraries

We don't have this right now.

znado:
@ckt624 can you figure out which versions of the additional pip dependencies this requires, and add them to `setup.py`?

ckt624:
> @ckt624 can you figure out which versions of the additional pip dependencies this requires, and add them to `setup.py`?

Added.

znado (approved):

PR #39, opened 2021-09-29
Docker

kamalkraj:
1) adc850f
    - Added docker with Nvidia base image `CUDA11.1` + requirements with pinned versions.

2) 0880251
     - Updated setup.py: pkgs pinned to specific versions
     - Especially all JAX libs are getting frequent releases with breaking changes, it is better to use pinned version for development.
     - Now both PyTorch and JAX is will run using CUDA 11.1 
     - Updated minimum required python to 3.7. python 3.6 contextlib don't have  `nullcontext` https://stackoverflow.com/questions/45187286/how-do-i-write-a-null-no-op-contextmanager-in-python
       jaxlib build for CUDA 11.1 onwards not available in py36.

 3) f513174
     - updated readme with docker instructions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+FROM nvidia/cuda:11.1.1-cudnn8-devel-ubuntu20.04
+
+RUN apt-get update && \
+    apt-get install -y bash \
+                   build-essential \
+                   git \
+                   curl \
+                   ca-certificates \
+                   python3 \
+                   python3-pip && \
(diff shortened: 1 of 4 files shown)


znado:
This is awesome, thanks so much for setting this up!

Do we need both a `requirements.txt` and a `setup.py`, or can we just keep the `setup.py` and run `pip install .`?

kamalkraj:
@znado 
Updated docker to install from setup.py

PR #46, opened 2022-01-14
Librispeech readme update

kamalkraj:
Update readme with complete run instructions 

@fsschneider @znado @ckt624


workloads/librispeech/prepare_data.py
```diff
@@ -111,7 +111,7 @@ def extract_spect_mvn(audio_path):
 if __name__ == '__main__':
   data_dir = sys.argv[1]
 
-  trans_dir = os.getcwd() + 'data'
+  trans_dir = os.getcwd() + '/data'
```
znado, line 114: for both of these, can we instead use `os.path.join` (imported via `import os.path`)?

workloads/librispeech/README.md
```diff
@@ -8,22 +8,38 @@ and evaluate it on Librispeech test clean dataset. We get Character Error Rate
...
+# [Optional] Delete the tar.gz files to save disk space
+# rm *.tar.gz
 
 cd ..
 mkdir work_dir
 cd work_dir
 
-python prepare_data.py data_dir/LibriSpeech
+# Run preprocessing script
+python ../prepare_data.py ../data_dir/LibriSpeech/
```
znado, line 28: are the `../` necessary? aren't those scripts in this dir?

znado: same with above `../download_data.sh`

znado (approved):

fsschneider:
@kamalkraj If you address the comments left by @znado, then I can merge this PR.

workloads/librispeech/prepare_data.py
kamalkraj: done

workloads/librispeech/README.md
kamalkraj: The `../` is used because both the `download_data.sh` and `prepare_data.py` is coded to store the data in the dir from which it runs. 
`download.sh` is run from `data_dir` and that's where the data get downloaded and unzipped.

I can update both scripts take path argument and solve this. 

Let me know your thoughts on this.

znado:
closing because this is now outdated

PR #47, opened 2022-01-17
Librispeech JAX workload

kamalkraj:
- [x] JAX CTC loss already added by zack 
- [x] add model
- [x] add input pipeline with fixed padding
- [x] update and parameterise preprocessing script for both PyTorch and JAX
- [x] add model training
    - [x] add training loop
    - [x] make model work with jit 
    - [x] fix loss dynamic mask creation
- [x] add model validation

### DIFF
--- a/workloads/librispeech/librispeech_jax/input_pipeline.py
+++ b/workloads/librispeech/librispeech_jax/input_pipeline.py
@@ -1,0 +1,57 @@
+"""Data pipeline for LibriSpeech dataset modified from https://github.com/lsari/librispeech_100/blob/main/dataset.py."""
+
+import json
+
+import jax.numpy as jnp
+import numpy as np
+import pandas as pd
+import torch
+
+
+class LibriSpeechDataset(torch.utils.data.Dataset):
+
+  def __init__(self, feat_csv):
+    self.df = pd.read_csv(feat_csv)
+    self.sample_size = len(self.df)
+    self.df["id"] = list(range(self.sample_size))
+
+  def __getitem__(self, idx):
+    sample = self.df.iloc[idx]
+    trn_ids = json.loads(sample["trans_ids"])
+
+    feature = np.load(sample["features"])
+    index = sample["id"]
+    return index, feature, trn_ids
+
+  def __len__(self):
+    return self.sample_size
+
+  def pad_collate(self, batch):
+    max_input_len = 2453
+    max_target_len = 405
+
+    for elem in batch:
+      index, feature, trn = elem
+      max_input_len = max_input_len if max_input_len > feature.shape[
+          0] else feature.shape[0]
+      max_target_len = max_target_len if max_target_len > len(trn) else len(trn)
+
+    for i, elem in enumerate(batch):
+      index, f, trn = elem
+      input_length = np.array(f.shape[0])
+      input_dim = f.shape[1]
+      feature = np.zeros((max_input_len, input_dim), dtype=np.float)
+      feature[:f.shape[0], :f.shape[1]] = f
+      trn = np.pad(
+          trn, (0, max_target_len - len(trn)), "constant", constant_values=0)
+
+      batch[i] = (int(index), feature, trn, input_length)
+
+    batch.sort(key=lambda x: x[3], reverse=True)
+
+    index = np.array([x[0] for x in batch], dtype=jnp.int32)
+    feature = np.array([x[1] for x in batch], dtype=jnp.float32)
+    trn = np.array([x[2] for x in batch], dtype=jnp.float32)
+    input_length = np.array([x[3] for x in batch], dtype=jnp.int32)
+
+    return index, feature, trn, input_length
--- a/workloads/librispeech/librispeech_jax/input_pipeline.py
+++ b/workloads/librispeech/librispeech_jax/input_pipeline.py
@@ -30,28 +30,26 @@ class LibriSpeechDataset(torch.utils.data.Dataset):
     max_input_len = 2453
     max_target_len = 405
 
-    for elem in batch:
-      index, feature, trn = elem
-      max_input_len = max_input_len if max_input_len > feature.shape[
-          0] else feature.shape[0]
-      max_target_len = max_target_len if max_target_len > len(trn) else len(trn)
-
     for i, elem in enumerate(batch):
       index, f, trn = elem
       input_length = np.array(f.shape[0])
       input_dim = f.shape[1]
       feature = np.zeros((max_input_len, input_dim), dtype=np.float)
       feature[:f.shape[0], :f.shape[1]] = f
+      target_len = len(trn)
+      target_padding = np.zeros(max_target_len)
+      target_padding[target_len:] = 1 
       trn = np.pad(
           trn, (0, max_target_len - len(trn)), "constant", constant_values=0)
 
-      batch[i] = (int(index), feature, trn, input_length)
+      batch[i] = (int(index), feature, trn, input_length, target_padding)
 
     batch.sort(key=lambda x: x[3], reverse=True)
 
     index = np.array([x[0] for x in batch], dtype=jnp.int32)
     feature = np.array([x[1] for x in batch], dtype=jnp.float32)
-    trn = np.array([x[2] for x in batch], dtype=jnp.float32)
+    trn = np.array([x[2] for x in batch], dtype=jnp.int32)
     input_length = np.array([x[3] for x in batch], dtype=jnp.int32)
+    target_padding = np.array([x[4] for x in batch], dtype=jnp.int32)
 
-    return index, feature, trn, input_length
+    return index, feature, trn, input_length, target_padding
--- a/workloads/librispeech/librispeech_jax/models.py
+++ b/workloads/librispeech/librispeech_jax/models.py
@@ -9,15 +9,6 @@ import jax.numpy as jnp
 from flax import linen as nn
 
 
-class Sequential(nn.Module):
-  layers: Sequence[nn.Module]
-
-  def __call__(self, x):
-    for layer in self.layers:
-      x = layer(x)
-    return x
-
-
 class SequenceWise(nn.Module):
   """Collapses input of dim T*N*H to (T*N)*H, and applies to a module.
 
@@ -58,7 +49,7 @@ class MaskConv(nn.Module):
       mask = mask.astype(jnp.float32)
       x *= mask
     x = x.transpose(0, 3, 1, 2)
-    return x, lengths
+    return x
   
   
 @jax.vmap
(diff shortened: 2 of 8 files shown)


workloads/librispeech/librispeech_jax/input_pipeline.py
```diff
@@ -0,0 +1,55 @@
+"""Data pipeline for LibriSpeech dataset modified from https://github.com/lsari/librispeech_100/blob/main/dataset.py."""
```
znado, line 1: it looks like this file is 90% the same as the pytorch version; is it possible to move them into a common `workloads/librispeech/input_pipeline.py`, and then subclass one here where the only thing it does is:
- override pad_collate
- call the super class impl
- do the final few `np.array` lines

and then another subclass in the pytorch/ dir which does the same thing but just calls `return torch.utils.data.dataloader.default_collate(batch)`

workloads/librispeech/librispeech_jax/input_pipeline.py
```diff
@@ -0,0 +1,55 @@
...
+
+import json
+
+import jax.numpy as jnp
+import numpy as np
+import pandas as pd
+import torch
+
+
+class LibriSpeechDataset(torch.utils.data.Dataset):
```
znado, line 11: now that we are requiring pytorch be installed for jax workloads (which is fine), we should change `setup.py` to always install pytorch

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+    inputs = inputs.transpose(1, 0, 2)
+    inputs = self.rnn(inputs, lengths)
+    inputs = inputs.transpose(1, 0, 2)
+    # (TxNxH*2) -> (TxNxH) by sum
+    inputs = inputs.reshape(inputs.shape[0], inputs.shape[1], -1).sum(2)
+    return inputs
+
+
+def hard_tanh(x, min_value=-1., max_value=1.):
+
```
znado, line 165: delete unnecessary newline

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+    self.fc = SequenceWise(fully_connected)
+
+  def __call__(self, inputs, lengths, training=False):
+    output_lengths = get_seq_lens(lengths, self.conv.seq_module)
+
+    x, _ = self.conv(inputs, lengths)
+
+    sizes = x.shape
+    x = x.reshape(sizes[0], sizes[1] * sizes[2],
+               sizes[3])
```
znado, line 248: can all of these `sizes` be on the same line?

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+      nn.BatchNorm(use_running_average=True),
+      nn.Dense(self.num_classes, use_bias=False)
+    ])
+
+    self.fc = SequenceWise(fully_connected)
+
+  def __call__(self, inputs, lengths, training=False):
+    output_lengths = get_seq_lens(lengths, self.conv.seq_module)
+
+    x, _ = self.conv(inputs, lengths)
```
znado, line 244: if we never use the second return value from `MaskConv`, can we remove it?

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+
+    Args:
+      input_length: 1D Tensor
+
+    Returns:
+      1D Tensor scaled by model
+    """
+    seq_len = input_length
+    for m in conv_seq_module:
+      if isinstance(m, nn.Conv):
```
znado, line 180: is there ever an instance where this is false?

workloads/librispeech/librispeech_jax/submission.py
```diff
@@ -0,0 +1,156 @@
...
+        strides=(2, 2),
+        padding=((20, 20), (5, 5)),
+      ),
+      nn.Conv(
+        features=32,
+        kernel_size=(21, 11),
+        strides=(2, 1),
+        padding=((10, 10), (5, 5)),
+      )
+    ]
```
znado, line 116: do we need to write out nn.Conv modules for this? can't we just provide the following to the function (hardcoded) instead of needing to import nn in this file?
- padding[1][1]
- kernel_dilation,
- kernel_size[1]
- strides[1]

workloads/librispeech/librispeech_jax/submission.py
```diff
@@ -0,0 +1,156 @@
...
+    loss_type: spec.LossType,
+    optimizer_state: spec.OptimizerState,
+    eval_results: List[Tuple[int, float]],
+    global_step: int,
+    rng: spec.RandomState) -> spec.UpdateReturn:
+  """Return (updated_optimizer_state, updated_params, updated_model_state)."""
+  _, features, transcripts, input_lengths,  transcripts_padding = input_batch
+  features = jnp.expand_dims(features.transpose(0, 2, 1), axis=1)
+  num_devices = jax.local_device_count()
+  reshaped_features = jnp.reshape(
```
znado, line 89: I think that all of this reshaping, padding, etc. should be done in the input pipeline, because it should be the same regardless of the optimizer (and really `submission.py` should only define new code relating to the optimizer)

workloads/librispeech/librispeech_jax/workload.py
```diff
@@ -0,0 +1,286 @@
...
+  def model_params_types(self):
+    pass
+
+  @property
+  def max_allowed_runtime_sec(self):
+    return 80000
+
+  @property
+  def eval_period_time_sec(self):
+    return 800
```
znado, line 144: all of these properties should be constant between pytorch and jax, would it be possible for you to make a parent `LibriSpeechWorkload` abstract class in `workloads/librispeech` that defines these, and then have both the Jax and Pytorch subclasses inherit from it?

workloads/librispeech/librispeech_jax/workload.py
```diff
@@ -0,0 +1,286 @@
...
+      features = jnp.expand_dims(features.transpose(0, 2, 1), axis=1)
+      reshaped_features = jnp.reshape(
+      features,
+      (num_devices, features.shape[0] // num_devices, *features.shape[1:]))
+      reshaped_input_lengths = jnp.reshape(
+          input_lengths,
+          (num_devices, input_lengths.shape[0] // num_devices, *input_lengths.shape[1:]))
+      batch = {
+        'input': (reshaped_features, reshaped_input_lengths)
+      }
```
znado, line 267: if you move the reshaping to input_pipeline.py, can this also be removed?

znado (changes requested):
Great start! A few points I think we can address and then it should be good to go after another round of review.

danielsnider:
Hi @kamalkraj,
I want to do some held-out workload experiments for LibriSpeech, preferrably in JAX. Does this pull request currently converge to the target accuracy? I don't see any learning improvement after 12 hours for JAX. It starts at .99 and stays at .99 accuracy. While when I run the PyTorch LibriSpeech workload, it starts at .99 and after a few hours it is approaching 0.2 accuracy. 

This is how I run the Jax and PyTorch workloads. Please let me know if I'm making a mistake, or if the JAX implementation is not yet tuned to converge.
```
# For Jax
python3 submission_runner.py --workload=librispeech_jax --submission_path=workloads/librispeech/librispeech_jax/submission.py tuning_search_space=baselines/librispeech/tuning_search_space.json --data_dir=/FS1/dataset/librispeech100/work_dir/data

# For PyTorch
python3 algorithmic_efficiency/submission_runner.py --framework=pytorch --workload=librispeech_pytorch --submission_path=baselines/librispeech/librispeech_pytorch/submission.py --tuning_search_space=baselines/librispeech/tuning_search_space.json --data_dir=/FS1/dataset/librispeech100/work_dir/data
```

Thanks!

kamalkraj:
Hi @danielsnider,

I haven't really trained this JAX model for long to compare the scores, due to GPU resource constraints. I will try to address the @znado comments by this week and also try training the model

workloads/librispeech/prepare_data.py
```diff
@@ -123,7 +140,7 @@ def extract_spect_mvn(audio_path):
   ]
   for subset in subset_list:
     print(subset)
-    df = get_txt('{}/{}'.format(data_dir, subset), output_label_dict)
+    df = get_txt('{}/{}'.format(data_dir, subset), output_label_dict,add_additional_blank=add_additional_blank)
```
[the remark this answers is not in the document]
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+import functools
+import math
+from typing import Sequence
+
+import jax
+import jax.numpy as jnp
+from flax import linen as nn
+
+
+class Sequential(nn.Module):
```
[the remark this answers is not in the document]
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+class CNNLSTM(nn.Module):
+
+  num_classes = 29
+  hidden_size = 768
+  hidden_layers = 5
+  context = 20
+
+  def setup(self):
```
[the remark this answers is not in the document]
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+          hidden_size=self.hidden_size
+          )
+      rnns.append(rnn)
+
+    self.rnns = rnns
+
+    fully_connected = Sequential([
+      nn.BatchNorm(use_running_average=True),
+      nn.Dense(self.num_classes, use_bias=False)
+    ])
```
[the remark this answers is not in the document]
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
```diff
@@ -0,0 +1,257 @@
...
+    # Use fixed random key since default state init fn is just zeros.
+    return nn.OptimizedLSTMCell.initialize_carry(
+        jax.random.PRNGKey(0), batch_dims, hidden_size)
+
+
+class SimpleBiLSTM(nn.Module):
+  """A simple bi-directional LSTM."""
+  hidden_size: int
+
+  def setup(self):
```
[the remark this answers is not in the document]
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
kamalkraj: done

workloads/librispeech/librispeech_jax/models.py
kamalkraj: yes

kamalkraj: `conv_seq_module` also contains `BatchNorm` and `hard_tanh`.

danielsnider:
By the way, when testing this branch the only thing I changed (in my own branch) was to skip a batch if it doesn't split equality across GPU devices, because I had 4 GPUs and the code assumes 8 GPUs. https://github.com/danielsnider/algorithmic-efficiency/commit/77d7d15db1ebb54cb3464166117c4f304e203ee3
I think it would be good to support something like this.

znado:
Closing as outdated.

PR #51, opened 2022-01-24
Restructuring, linting, testing

fsschneider:
This PR changed the following:

- Created a main library directory (`algorithmic_efficiency`) which includes the base files such as `submission_runner.py` and the `workloads` directory.
- Separated the main library (now in the `algorithmic_efficiency` directory) from submissions (now in the `sample_submissions` directory).
- Moved all the main package configuration (including config for `pylint`, `isort`, `yapf`, and `pytes`) into `setup.cfg`.
- Include Github Actions for linting via `pylint`, `isort`, `yapf` and testing via `pytest`.
  - Fix `isort` errors.
  - Fix `yapf` errors.
  - Create a simple exemplary test.
- Add a `__version__` attribute, currently set to `0.0.1`.
- Switch to absolute imports.
- Add issue template for `new_workloads`, `feature_requests` and `bug_reports`.
- Updated `README.md` with current commands for running a workload (accounts for new repository structure)
- Add badges to show linting and testing status, license, and coding style.

@znado feel free to change the configs of the linting to suit whatever you are used to.


.github/workflows/linting.yml
```diff
@@ -0,0 +1,54 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install yapf
+      run: |
+        python -m pip install --upgrade pip
+        pip install yapf
+    - name: Run yapf
+      run: |
+        yapf . --diff --recursive
```
znado, line 54: can we add `--style=yapf` to this? I think that is the style that is most similar to our internal formatter

znado (approved):
Overall LGTM, one comment is what do you think about renaming `sample_submissions/` to `baselines/`? Then we can put all the baseline submissions in there too

runame (approved):
It also LGTM. Maybe the remaining pylint errors should already be addressed in this PR?

.github/workflows/linting.yml
fsschneider: The `yapf` configuration sits in the `setup.cfg` file (https://github.com/fsschneider/algorithmic-efficiency/blob/ada14168b738c99f640508f6a7959945c3e939c4/setup.cfg#L82-L86). As far as I know, yapf automatically uses this configuration so adding `--style=yapf` to it here is unnecessary. Using the configuration in `setup.cfg` has the advantage of one central place to define our coding style. Running it locally, either via the command line or in an editor should by default also use this configuration.

Originally, I was using the `google` style, with `column_limit=80` and `indent_width=2`. Changed it now to `yapf` style without any further modifications.

PR #52, opened 2022-01-26
Fixing ctcdecode version (latest seems to be 1.0.2 https://pypi.org/project/ctcdecode/) and fixing jax install on GPUs

znado:


fsschneider:
I ran into issues installing `ctcdecode` from PyPi.
Even a simple `pip install ctcdecode` failed with `ERROR: Failed building wheel for ctcdecode`. Pip install issues seem to be quite common with `ctcdecode`, see for example https://github.com/parlance/ctcdecode/issues/186, https://github.com/parlance/ctcdecode/issues/182, or https://github.com/parlance/ctcdecode/issues/166.

What worked for me was to install it directly from Github, i.e. `pip install git+https://github.com/parlance/ctcdecode.git`. This does install version `1.0.3`. Unfortunately, this version is not tagged on Github.

Is anyone else running into these issues? We could fix a commit and set our install in the `setup.py` to this commit. However, if I remember correctly, this would prevent us from publishing our package on PyPi.

Another note, I think that `ctcdecode` is currently only used for the PyTorch version. So I would vote to move both `ctcdecode` and `levenshtein` to the `options.extra_require`. Perhaps, we need new sections there for `pytorch_requirements` and `jax_requirements`. The `pytorch` options would then install `torch` but also the `pytorch_requirements`.

I think this would be a cleaner solution, instead of requiring everyone to install packages that are framework-specific. `ctcdecode`, for example, requires `torch` itself. So JAX users would be forced to install `torch` as well.

kamalkraj:
Hi @fsschneider,
The JAX version also uses `ctcdecode` for inference. To install `ctcdecode` properly you should have `torch` installed in the `venv`. 

I think we should have `requirements.txt` specific to workloads in their corresponding folders. In the main `setup.py` we can keep the only `torch`, `jax` , `tensorflow` pkgs alone. As the library grows it will be difficult to keep dependencies for all the workload in a single setup and the users will end up installing all the pkgs for running even a single workload.

znado:
For development purposes I think it makes sense to split out `setup.py` into having a different key in `extras_require` for each workload, but when we run the actually competition we'll need to install them all at once, so we'll need to have a valid install configuration including them all at some point.

I also recently had to comment out the `ctcdecode` install because even when I already had torch installed in the virtualenv, it failed with some `gcc` error.

Also, given that for the librispeech workload we have a pytorch dataloader for the jax workload (which is great, the more shared code the better!) we'll need to install some (all?) of the torch deps for jax too.

fsschneider:
Superseded by #55

PR #55, opened 2022-02-08
Organize dependencies & some clean up

### LINKED ISSUE #19: Only import selected workload
Currently, the `submission_runner` takes the `WORKLOADS` dictionary, e.g.:
https://github.com/mlcommons/algorithmic-efficiency/blob/584769427aceeac2835c055fae40f159d4a415d7/submission_runner.py#L30-L33

And attempts to import every module, replacing the dictionary values of `WORKLOADS` with the workload class object:

https://github.com/mlcommons/algorithmic-efficiency/blob/584769427aceeac2835c055fae40f159d4a415d7/submission_runner.py#L292-L297

In case of import failures however, the dictionary values are unchanged, and if we attempt to run that workload, we get a less instructive error:
```
AttributeError: 'dict' object has no attribute 'build_input_queue'
```

Moreover, since most users will be submitting in one framework, the log will be polluted with many import errors for the other frameworks.

Instead, perhaps we only import the workload selected by `FLAGS.workload`, and raise an `ImportError` if fails. 

cc: @znado ?

fsschneider:
This PR performs some general cleanup and allows workload-specific dependencies:

- Fixes JAX install according to #52.
- Introduces workload-specific dependencies which can be installed e.g. via `pip install -e '.[librispeech]'`. There is also the option to use `[full]` to install dependencies for all workloads.
- Workloads are now defined without appending the framework, e.g. `--workload=mnist_jax` should now be `--workload=mnist`. It automatically combines the `workload` and `framework` arguments. This provides a clearer interface as previously one could run with the options `--framework=jax --workload=mnist_pytorch` as discussed in #19. This PR closes #19.
- Fixes some `pylint` errors in `baselines/` directory.

### DIFF
--- a/algorithmic_efficiency/workloads/imagenet/imagenet_jax/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet/imagenet_jax/workload.py
@@ -1,8 +1,8 @@
 """ImageNet workload implemented in Jax.
 
 python3 submission_runner.py \
-    --workload=imagenet_jax \
-    --submission_path=workloads/imagenet/imagenet_jax/submission.py \
+    --workload=imagenet \
+    --submission_path=baselines/imagenet/imagenet_jax/submission.py \
     --num_tuning_trials=1
 """
 import functools
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,9 +36,6 @@ setup_requires =
 # Dependencies of the project:
 install_requires =
   absl-py==0.14.0
-  # ctcdecode>=1.0.3
-  levenshtein>=0.12.0
-  librosa>=0.8.1
   numpy>=1.19.2
   pandas>=1.3.1
   six>=1.15.0
@@ -55,12 +52,34 @@ python_requires = >=3.7
 # Use as `pip install -e '.[jax-gpu]' -f https://storage.googleapis.com/jax-releases/jax_releases.html`
 # or `pip install -e '.[dev]'`
 
+# Bundled installs #
+
+# All workloads
+full =
+  %(librispeech)s
+
+# All workloads plus development dependencies
+full-dev =
+  %(full)s
+  %(dev)s
+
 # Dependencies for developing the package
 dev =
   isort
   pylint
   pytest
   yapf
+
+
+# Workloads #
+
+librispeech = 
+  ctcdecode>=1.0.2
+  levenshtein>=0.12.0
+  librosa>=0.8.1
+
+
+# Frameworks #
 
 # JAX Core
 jax_core_deps =
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,7 @@ full =
   %(librispeech)s
 
 # All workloads plus development dependencies
-full-dev =
+full_dev =
   %(full)s
   %(dev)s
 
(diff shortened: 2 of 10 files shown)


setup.cfg
```diff
@@ -34,134 +36,142 @@ setup_requires =
...
 # JAX GPU
 jax_gpu =
   %(jax_core_deps)s
-  jaxlib==0.1.71+cuda111
+  jax[cuda]==0.2.28
+  jaxlib==0.1.76+cuda11.cudnn82
 
 # PyTorch
 pytorch =
   torch==1.9.1+cu111
```
runame, line 108: Is there a reason why we don't use the latest stable release (1.10.2)?

runame (approved):
Looks good! One general comment: the string style (`"` vs `'`) is inconsistent, I think it's worth fixing this at some point. My vote is in favor of `'`.

algorithmic_efficiency/workloads/imagenet/imagenet_jax/workload.py
```diff
@@ -1,34 +1,30 @@
...
 import functools
 from typing import Tuple
 
-import optax
-import tensorflow as tf
-import tensorflow_datasets as tfds
-
-# Hide any GPUs form TensorFlow. Otherwise TF might reserve memory and make it
-# unavailable to JAX.
-tf.config.experimental.set_visible_devices([], 'GPU')
```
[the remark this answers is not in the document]
fsschneider: I believe this is already set in our `submission_runner.py`, see:
https://github.com/mlcommons/algorithmic-efficiency/blob/cc58b2353d34d28f20e393ef398cc912c43726cc/algorithmic_efficiency/submission_runner.py#L282-L287

As this is set every time `framework=='jax'` is used, I don't think it is necessary to set this again for the `imagenet_jax` workload.

setup.cfg
fsschneider: I don't think there is one. Perhaps this was the stable release at the time of initializing the package.

I also noticed that the learning rate schedule in `imagenet_pytorch` requires `torch>=1.10`. We will probably have to discuss and fix the package versions close to the Call for Submissions.

I am happy to bump the version for now. @runame this would mean `torch==1.10.2+cu111`?

algorithmic_efficiency/workloads/imagenet/imagenet_jax/workload.py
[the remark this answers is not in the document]
runame: I see, that should indeed be sufficient, thanks.

setup.cfg
runame: Ok, the current version is `torch==1.10.2+cu113` and `torchvision==0.11.3+cu113`. Hopefully this doesn't break anything, but most of the changes should be backwards compatible.

fsschneider:
> Looks good! One general comment: the string style (" vs ') is inconsistent, I think it's worth fixing this at some point. My vote is in favor of '.

True. It does bug me a little bit but not yet enough for me to take the time to change it :) I agree though that it should be on the list and I am fine with `'`.

One other thing I noted: Currently even the `mnist` workload is broken by the changed I introduced. This is due to changing `param_shapes` to be a property. I think I should fix this, before we merge this PR.

Please correct me if I am wrong, but I think our current idea for `param_shapes` and `model_params_types` is that they are both property of the workload and should be defined workload-agnostic. For example, `param_shapes` could be a list of lists with shapes of the parameters. It would be defined in the workload-agnostic `workload.py` file (the parent class for the pytorch and jax versions) and we could write a unittest to check whether both framework-specfic implementations produce a model with these exact `param_shapes`. Is that right?

PR #58, opened 2022-03-01
Graph workload in Jax

znado:
Reaches 0.24 mAP in 1671 seconds on a 96 core CPU, 8 V100 system with a per-GPU batch size of 256 (for 2048 total batch size). Have not tried to further tune the baseline, or see if we could get an even higher mAP by training longer.

Originally built off the excellent work by runame@.

Also adding a bit of formatting config that is nicer, and moving `submission_runner.py` back into the root dir for easier python importing.


algorithmic_efficiency/workloads/ogb/ogb_jax/input_pipeline.py
```diff
@@ -0,0 +1,161 @@
...
+    num_shards: How many devices we should be able to shard the batch into.
+  Yields:
+    Batch in the init2winit format. Each field is a list of num_shards separate
+    smaller batches.
+  """
+  if not num_shards:
+    num_shards = jax.device_count()
+
+  # We will construct num_shards smaller batches and then put them together.
+  batch_size /= num_shards
```
runame, line 108: It looks like `batch_size` is here assumed to be the global batch size and not the per-device one?

algorithmic_efficiency/workloads/ogb/ogb_jax/workload.py
```diff
@@ -0,0 +1,211 @@
...
+  # Does NOT apply regularization, which is left to the submitter to do in
+  # `update_params`.
+  def loss_fn(
+      self,
+      label_batch: spec.Tensor,
+      logits_batch: spec.Tensor,
+      mask_batch: Optional[spec.Tensor]) -> spec.Tensor:  # differentiable
+    loss = self._binary_cross_entropy_with_mask(
+        labels=label_batch, logits=logits_batch, mask=mask_batch)
+    mean_loss = jnp.sum(jnp.where(mask_batch, loss, 0)) / jnp.sum(mask_batch)
```
runame, line 155: Shouldn't the loss function return the per-example instead of the mean loss?

algorithmic_efficiency/workloads/ogb/ogb_jax/workload.py
```diff
@@ -0,0 +1,211 @@
...
+    return mean_loss
+
+  def _eval_metric(self, labels, logits, masks):
+    loss = self.loss_fn(labels, logits, masks)
```
runame, line 159: This returns the mean loss but I think it should be per-example loss.

algorithmic_efficiency/workloads/ogb/ogb_jax/workload.py
```diff
@@ -0,0 +1,211 @@
...
+                 params: spec.ParameterContainer,
+                 model_state: spec.ModelAuxiliaryState,
+                 rng: spec.RandomState,
+                 data_dir: str):
+    """Run a full evaluation of the model."""
+    data_rng, model_rng = prng.split(rng, 2)
+    eval_per_core_batch_size = 1024
+    if self._eval_iterator is None:
+      self._eval_iterator = self._build_iterator(
+          data_rng, 'validation', data_dir, batch_size=eval_per_core_batch_size)
```
runame, line 188: Do we want to use the `validation` or `test` split here? For the validation split the mAP is >25 in the Flax example (25.2 without and 25.7 with early stopping).

baselines/ogb/ogb_jax/submission.py
```diff
@@ -0,0 +1,118 @@
...
+  def loss_fn(params):
+    logits_batch, new_model_state  = workload.model_fn(
+        params,
+        input_batch,
+        model_state,
+        spec.ForwardPassMode.TRAIN,
+        rng,
+        update_batch_norm=True)
+    loss = workload.loss_fn(label_batch, logits_batch, mask_batch)
+    mean_loss = jnp.sum(jnp.where(mask_batch, loss, 0)) / jnp.sum(mask_batch)
```
runame, line 56: In the current implementation `workload.loss_fn()` seems to already return the mean loss.

runame (commented):
Looks good to me! Thanks for fixing the pmap issue.

I only have questions regarding the `batch_size` and `workload.loss_fn()`, see my comments below.

baselines/ogb/ogb_jax/submission.py
```diff
@@ -0,0 +1,118 @@
...
+import jax
+from jax import lax
+import jax.numpy as jnp
+import optax
+
+from algorithmic_efficiency import spec
+
+
+def get_batch_size(workload_name):
+  # Return the per-device batch size.
```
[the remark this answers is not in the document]
znado: Ah good point, I like that convention more anyways. Changed.

algorithmic_efficiency/workloads/ogb/ogb_jax/workload.py
znado: We should probably run on both, and append `split/` to make the keys `split/metric_name`. I'll do that in a follow-up CL for all workloads.

algorithmic_efficiency/workloads/ogb/ogb_jax/workload.py
znado: Yep! Thanks for catching that. Done.

baselines/ogb/ogb_jax/submission.py
znado: Fixed.

znado:
Thanks so much for the thorough review!

I renamed OGB to OGBG like the dataset, fixed and documented returning the global batch size, and return per-example losses in OGBG.

runame (approved):
Thanks for addressing the comments! I have noticed that in some workloads the batch size is also just the per-device batch size, e.g. in the Jax WMT workload, see [this line](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py#L291). But I will just fix this in the PyTorch WMT PR.

algorithmic_efficiency/workloads/ogb/ogb_jax/input_pipeline.py
znado: So given that we accumulate batches until we hit num_shards batches (`if count == num_shards:`), we should have been running with a per-device batch size of 256, for a total of 2048.

runame: Ah I see now, thanks!

PR #59, opened 2022-03-02
Fixing lint and formatting.

znado:
Ran yapf and pylint and fixed the resulting issues.

### DIFF
--- a/algorithmic_efficiency/spec.py
+++ b/algorithmic_efficiency/spec.py
@@ -277,5 +277,5 @@ def data_selection(workload: Workload,
 
 
 def get_batch_size(workload_name):
-  """Return a batch size to use for a given workload."""
+  """Return a per-device batch size to use for a given workload."""
   pass
--- a/algorithmic_efficiency/spec.py
+++ b/algorithmic_efficiency/spec.py
@@ -138,8 +138,13 @@ class Workload(metaclass=abc.ABCMeta):
 
   @property
   @abc.abstractmethod
-  def num_eval_examples(self):
-    """The size of the evaluation set."""
+  def num_eval_train_examples(self):
+    """The number of training examples to evaluate metrics on."""
+
+  @property
+  @abc.abstractmethod
+  def num_validation_examples(self):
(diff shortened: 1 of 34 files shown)


algorithmic_efficiency/spec.py
```diff
@@ -238,5 +280,5 @@ def data_selection(workload: Workload, input_queue: Iterator[Tuple[Tensor,
 
 
 def get_batch_size(workload_name):
-  """Return a batch size to use for a given workload."""
+  """Return a per-device batch size to use for a given workload."""
```
fsschneider, line 283: I think you now decided that `get_batch_size` should return the global batch size instead of the per-device one, no?

algorithmic_efficiency/spec.py
```diff
@@ -97,41 +119,54 @@ def param_shapes(self):
   def model_params_types(self) -> ParameterType:
```
fsschneider, line 119: I might be out of the loop here, but isn't `model_params_types` meant to be used similar to `param_shapes` just that it provides the type instead of the shapes? Shouldn't it then also be a `@property`?

fsschneider (approved):
Looks good to me!
I just left two small comments.

I saw that you moved the `submission_runner` out of the main package and moved the test files to the corresponding files. At the moment, I don't have a firm idea on how we want to distribute the code for the competition. Originally, I was envisioning serving it as a `pip install` package. Currently, this would not include the crucial `submission_runner` but would include all the test files that would be unnecessary for submitters.
This is not the content of this PR, but I am just curious to hear your opinion.

Let me know what you think about the other two comments and then I can merge :)

algorithmic_efficiency/spec.py
znado: Yep, missed that. Fixed.

algorithmic_efficiency/spec.py
znado: yep, this was out of sync with the GNN workload PR, I merged and updated.

PR #60, opened 2022-03-05
[Logging] Save information about training progress to disk

danielsnider:
Currently the only way that we log training accuracy and loss is to the console stdout. We would like a structured way to save the progress of training after every model eval to disk so that plots and analysis can be made. I welcome your suggestions on my design.

## Summary of Measurements Collected ([or full list here](https://github.com/UofT-EcoSystem/algorithmic-efficiency/tree/logging/baselines/logging_example#:~:text=Four%20files%20are%20written%20to%20the%20logging_dir%20folder%3A))
|                                               | `workload_`<br />`results.json` | `trial_`<br />`results.json`  | `eval_`<br />`results.csv` |
|-----------------------------------------------|-----------------------|---------------------|------------------|
| **Basic info** <br /> (ex. datetime)                     | ✔️                    | ✔️                  | ✔️               |
| **Extra metadata** <br /> (optional user-defined)        | ✔️                    | ✔️                  | ✔️               |
| **Workload info** <br /> (ex. target_value)              | ✔️                    | ✔️                  | ✔️               |
| **System software info** <br /> (ex. os_platform)        | ✔️                    | ✔️                    | ✔️                  |
| **System hardware info** <br /> (ex. gpu_model_name)     | ✔️                    |  ✔️                   |  ✔️                |
| **Workload score**                                | ✔️                    |                     |                  |
| **Tuning search space** <br /> (if used)                 | ✔️                    |                     |                  |
| **Trial hyperparameters**                         |                       | ✔️                  | ✔️               |
| **Hardware   utilization** <br /> (ex. gpu.avg.mem.util) |                       | ✔️                  | ✔️               |
| **Model eval metrics** <br /> (ex. loss)                 |                       | ✔️ (only<br /> last eval) | ✔️ (all evals)   |
| **Stopping   conditions** <br /> (ex. is_time_remaining) |                       | ✔️                  | ✔️               |
| **Step count**                                    |                       | ✔️*                  | ✔️              |
| **Accumulated time**                              |                       | ✔️*                  | ✔️              |

&ast; Caveat: When the tuning trial ran out of time before reaching the target value, one or more steps may have been completed since the last model evaluation. In this case the step count and accumulated time in `trial_results.json` will not exactly reflect when the model evaluation metrics were recorded. 

### Included in this PR:
- A new file called `logging_utils.py`.
- When `--log_dir [DIR]` is used on CLI with `submission_runner.py`, output files will be written to disk.
- Three files are written to disk:
  1. "metadata.json" is created at the start of a workload and it includes the
     datetime, workload name, and system configuration.
  2. "measurements.csv" is created for each hyperparameter tuning trial and a row is
     appended for every model evaluation. The information included is loss,
     accuracy, training step, time elapsed, hparams, workload properties,
     and hardware utilization.
  3. "packages.txt" is created at the start of a workload and it includes a list of the currently installed OS and python packages. However, this is disabled due to a security concern. The problem is that pip package list can contain a git repo url which can contain a users github personal access token.
- A new folder called `experiments` which includes a __[README.md](https://github.com/mlcommons/algorithmic-efficiency/blob/5ceb5ff0c3a75f30759275ceedb0a77638c0d927/experiments/README.md)__ with more detailed usage instructions and an example of measuring a loss curve and plotting it. __Please read this [README.md](https://github.com/mlcommons/algorithmic-efficiency/blob/5ceb5ff0c3a75f30759275ceedb0a77638c0d927/experiments/README.md) as it explains more about this PR.__ I welcome suggestions for a different location of this folder, or to not included at all.
- Checked with yapf and isort.
- I fixed as many pylint issues as I could. Some warnings remain but they are preferred or necessary.
- Includes typing hints and docstrings.

### Usage:

Run tests:
```pytest ./algorithmic_efficiency/logging_utils_test.py```

Run MNIST with logging: 
```
    --framework=jax \
    --workload=mnist \
    --submission_path=baselines/mnist/mnist_jax/submission.py \
    --tuning_search_space=baselines/mnist/tuning_search_space.json \
    --num_tuning_trials=1 \
    --eval_frequency_override='50 step' \
    --logging_dir='./logs'
```

### Why was the MLCommons Logging library not used?
After speaking to a developer of the [mlcommons/logging](https://github.com/mlcommons/logging) library, I found it to be unsuitable to our needs. This library is designed for the training and hpc benchmarks ([example usage](https://github.com/mlcommons/training_results_v1.1/blob/main/NVIDIA/benchmarks/bert/implementations/pytorch/run_pretraining.py#L1378)). It produces a single log in a JSON format that primarily records the start and stop time of training ([example log](https://github.com/mlcommons/training_results_v1.1/blob/main/NVIDIA/results/dgxa100_n128_ngc21.09_pytorch/bert/result_0.txt)), but we want to record more than training runtime. We are interested in training dynamics so we want to additionally log training loss, step count, hyperparameters, details about the workload, and more. They do have an interesting file called [constants.py](https://github.com/mlcommons/logging/blob/master/mlperf_logging/mllog/constants.py) which includes standard names for hyperparameters and optimizers, but at this point in my opinion it is not necessary or easy to build on this for our purposes.

### CODE REVIEW

**Comment by danielsnider** [comment]
I have some additional improvement plans, inspired by the [batch_science](https://github.com/google-research/google-research/tree/master/batch_science#description-of-the-data) repo, that I will implement after getting some feedback. I welcome feedback on plans:
- metadata.json should include the hyperparameter config if an external `tuning.json` was given. The batch_science repo uses a "parameter_configs" key for this.
- Rename metrics.csv to measurements.csv, as per the batch_science repo 
- Add  "status" key  to measurements.csv and metadata.json to be one of:
		- COMPLETE: If the trial was completed.
		- INCOMPLETE: If the trial was not completed for some reason (these trials can usually be ignored).
		- INFEASIBLE: If training diverged at any point.
- Improve the documentation to look more like the batch_science repo [here](https://github.com/google-research/google-research/tree/master/batch_science#description-of-the-data).

**Comment by danielsnider** [comment]
Question: Are there train/eval measurements that are missing? There is a long list of [Metrics and scoring for model evaluation in sklearn](https://scikit-learn.org/stable/modules/model_evaluation.html), but are eval metrics are always workload specific?

File: algorithmic_efficiency/logging_utils.py [hunk 1]
```diff
@@ -0,0 +1,492 @@
+from datetime import datetime
```
znado, line 1: nit: can you add a brief (even if redundant/obvious) file docstring?

algorithmic_efficiency/logging_utils.py
```diff
@@ -0,0 +1,492 @@
...
+
+from absl import flags
+from absl import logging
+import GPUtil
+import pandas as pd
+import psutil
+
+from algorithmic_efficiency import spec
+
+FLAGS = flags.FLAGS
```
znado, line 18: in general I try to avoid using flags within any library code, and instead just collect them all in main() (or one function down the call stack) and then pass them around as arguments. if flags are directly read throughout library code, it can lead to non-obvious behavior where the reader is not aware of how flag values are always used.

it seems like the extra logging data and eval freq are the only flags used, so they could probably be passed into the constructor or a function call as arguments.

algorithmic_efficiency/logging_utils.py
```diff
@@ -0,0 +1,492 @@
...
+    return metadata
+  try:
+    for item in extra_metadata_string_list:
+      key, value = item.split("=")
+      metadata['extra.' + key] = value
+  except:
+    logging.error(
+        'Failed to parse extra_metadata CLI arguments. Please check your '
+        'command.')
+    raise
```
znado, line 161: it could be more helpful to move the try/except into the for loop, and then if you can't parse the error then raise an exception and include the item that was unable to be parsed in the exception message.

Also instead of logging and a naked `raise`, I would do something like `raise ValueError('Failed to par...')`

algorithmic_efficiency/logging_utils.py
```diff
@@ -0,0 +1,492 @@
...
+    self.workload_name = workload_name
+    self.workload = workload
+    self.logging_dir = logging_dir
+    self.submission_path = submission_path
+    self.tuning_ruleset = tuning_ruleset
+    self.tuning_search_space_path = tuning_search_space_path
+    self.num_tuning_trials = num_tuning_trials
+    self.status = 'INCOMPLETE'
+    self.last_epoch_evaluated = None
+    self.workload_log_dir = os.path.join(self.logging_dir, self.workload_name)
```
znado, line 245: I don't mind keeping them as is, but in general if you don't plan on accessing these properties from outside the class, their names should start with a `_`.

algorithmic_efficiency/logging_utils.py
```diff
@@ -0,0 +1,492 @@
...
+      f.write('\n\nOS Packages:\n')
+      f.write(os_package_list)
+
+  def _write_trial_metadata_file(
+      self, workload: spec.Workload,
+      hyperparameters: Optional[spec.Hyperparamters], trial_idx: int,
+      global_step: int, batch_size: int, latest_eval_result: dict,
+      global_start_time: float, accumulated_submission_time: float,
+      goal_reached: bool, is_time_remaining: bool, training_complete: bool,
+      early_stop: bool):
```
znado, line 349: I know writing it out it is long/annoying, but when defining functions I think it is much cleaner to have a single function parameter per line, instead of a variable number of them per line. that way when scanning argument lists to see if you missed anything, it is easier to match up the orderings. if the linter does not like that, let me know and I'll figure out a way to make it stop.

algorithmic_efficiency/logging_utils.py
```diff
@@ -0,0 +1,492 @@
...
+               tuning_search_space_path: Optional[str] = None,
+               num_tuning_trials: Optional[int] = None):
+    self.workload_name = workload_name
+    self.workload = workload
+    self.logging_dir = logging_dir
+    self.submission_path = submission_path
+    self.tuning_ruleset = tuning_ruleset
+    self.tuning_search_space_path = tuning_search_space_path
+    self.num_tuning_trials = num_tuning_trials
+    self.status = 'INCOMPLETE'
```
znado, line 243: instead of having a property, maybe it would make sense to pass status as an argument to `_write_workload_metadata_file()`?

algorithmic_efficiency/logging_utils.py
```diff
@@ -0,0 +1,492 @@
...
+      # avoid overlap with other possible kys in the measurements dict.
+      hparams_dict = {f'hparam.{k}': v for k, v in hparams_dict.items()}
+      measurements.update(hparams_dict)
+
+    # Record workload properties
+    workload_properties = _get_workload_properties(workload)
+    measurements.update(workload_properties)
+
+    if FLAGS.extra_metadata:
+      extra_metadata = _get_extra_metadata_as_dict(FLAGS.extra_metadata)
```
znado, line 423: do we also need to write the same extra metadata to the measurements file on each row? is it sufficient to write it to  `metadata.json`?

algorithmic_efficiency/logging_utils.py
```diff
@@ -0,0 +1,492 @@
...
+    the user wants to eval this step."""
+    if not FLAGS.eval_frequency_override:
+      return False
+
+    try:
+      freq, unit = FLAGS.eval_frequency_override.split(' ')
+      freq = int(freq)
+      assert (unit in ['epoch', 'step'])
+    except:
+      logging.error(
```
znado, line 469: Similar to above, instead of logging and raise, I would `raise ValueError('Failed to parse e...')`

experiments/simple_example_mnist_loss/.gitignore
```diff
@@ -0,0 +1 @@
+!plot_loss.png
```
znado, line 1: do we need this or can we rely on the global gitignore?

experiments/simple_example_mnist_loss/measure_loss.sh
```diff
@@ -0,0 +1,37 @@
+# This script will run an MNIST training workload and save measurements as CSV and metadata as JSON.
```
znado, line 1: this is a nice script, but I'm not entirely sure if we need to include it in the repo, given that your README and the self-documenting FLAGS and code seem to provide all the information in here (plus, users will likely have their own ways of stringing together their experiment running framework with our codebase).

algorithmic_efficiency/submission_runner.py
```diff
@@ -147,7 +172,8 @@ def train_once(workload: spec.Workload, batch_size: int, data_dir: str,
                update_params: spec.UpdateParamsFn,
                data_selection: spec.DataSelectionFn,
                hyperparameters: Optional[spec.Hyperparamters],
-               rng: spec.RandomState) -> Tuple[spec.Timing, spec.Steps]:
+               rng: spec.RandomState, record: Callable,
```
znado, line 175: nit: can you replace the space after `rng: spec.RandomState,` with a newline?

znado (commented):
This looks fantastic overall! The code was super straight forward and I think it records a lot of super useful information we'll need. One high-level comment is: instead of making an `experiments/` dir, can you put your example code in a new dir `baselines/logging_example`? I (very soon) plan on moving all the existing "baselines" in `baselines/` into `tests/`, because they are not really baselines that run across all workloads, they are really test submissions for a single workload that exercise all the workload-specific functionality. `baselines/` should really be a collection of training algorithms that run across all workloads

experiments/simple_example_mnist_loss/measure_loss.sh
danielsnider: Agreed! I removed the `measure_loss.sh` script, and I removed the `plot_loss.py` script for the same reason.

experiments/simple_example_mnist_loss/.gitignore
danielsnider: I put the plot_loss.png in my own public [gist](https://gist.github.com/danielsnider/73519752010e430af55380f9bf2ee653) and linked it in the README.md. If that is okay, it leaves us with two fewer files in the repo the image and this .gitignore file👍.

danielsnider:
> One high-level comment is: instead of making an `experiments/` dir, can you put your example code in a new dir `baselines/logging_example`? I (very soon) plan on moving all the existing "baselines" in `baselines/` into `tests/`, because they are not really baselines that run across all workloads, they are really test submissions for a single workload that exercise all the workload-specific functionality. `baselines/` should really be a collection of training algorithms that run across all workloads

Done. Now there is just a single file called `baselines/logging_example/README.md`. I deleted the other files that were wholly contained in the README.

algorithmic_efficiency/logging_utils.py
danielsnider: Done.

algorithmic_efficiency/logging_utils.py
danielsnider: Fixed.

algorithmic_efficiency/logging_utils.py
danielsnider: Fixed. Thanks!

algorithmic_efficiency/logging_utils.py
danielsnider: Done. Nice touch! I found a trick to make yapf respect this preference: put a trailing comma after the last arugment, for example:
```
def foo(a, b,):
```

Becomes:
```
def foo(
    a,
    b,
):
```

Tested to work with yapf default settings:
```
$ yapf -i algorithmic_efficiency/logging_utils.py
```

algorithmic_efficiency/logging_utils.py
danielsnider: I agree this is simpler. Thanks! Done.

algorithmic_efficiency/logging_utils.py
danielsnider: I'm thinking about this and a larger rework eliminate duplication. I'll reply with more details later.

algorithmic_efficiency/logging_utils.py
danielsnider: Done. Thanks!

algorithmic_efficiency/submission_runner.py
danielsnider: Done. Thanks!

algorithmic_efficiency/logging_utils.py
danielsnider: Done. Thanks!

algorithmic_efficiency/logging_utils.py
danielsnider: Yeah, this info gets duplicated. My thinking here is I'm trying to prioritize easy analysis.  The trade-off is that it saves users from needing to combine data from multiple files when doing analysis. I understand the other side of the trade-off is: information overload and no single source of truth. But I don't think those are severe concerns here. If you agree, then lets leave this in.

danielsnider: I decided not to reduce duplication because of my goal to prioritize ease of analysis, but I'm still open to changing this. See the "[Summary of Measurements Collected](https://github.com/mlcommons/algorithmic-efficiency/blob/37521bf9b6f5a6a1a80dc59db38237b6a8745401/baselines/logging_example/README.md#summary-of-measurements-collected)" table to better understand the current duplicating of data.

danielsnider:
Hi @znado,

Thank you so much for the feedback. There we a lot of helpful tips! I really enjoyed it.

Take a look again at the PR and if you're happy, can we merge it? 

Summary of major changes:
- Added [logging_utils_test.py](https://github.com/mlcommons/algorithmic-efficiency/pull/60/files#diff-f686507f476558bdf7a3c493b5174361f970c45564c0ffcafe91ce407136d216). Huzzah!
- Renamed output files to make more sense:
    - `metadata.json` --> `workload_results.json`
    - `trial_[n]/metadata.json` --> `trial_[n]/trial_results.json`
    - `trial_[n]/measurements.csv` --> `trial_[n]/eval_results.csv`
- Removed `experiments` folder, keeping only [README.md](https://github.com/mlcommons/algorithmic-efficiency/pull/60/files#diff-ab5c741bc0e60da96dc6a4ba01a4c3aaa97e1f942cf7ebb34bd993e4acbb3f13).
    - Added "[Summary of Measurements Collected](https://github.com/mlcommons/algorithmic-efficiency/blob/37521bf9b6f5a6a1a80dc59db38237b6a8745401/baselines/logging_example/README.md#summary-of-measurements-collected)" markdown table to README.md.
- Synced with the main banch
- All checks pass

znado:
we're taking some of this logging logic and reusing it in a new PR, so I'm closing this.

PR #61, opened 2022-03-10
Merging Zack and Frank linting

znado:
Adding changes from PR 55 onto PR 59.


test_version.py
```diff
@@ -0,0 +1,13 @@
+"""Check whether the __version__ attribute is set correctly."""
```
fsschneider, line 1: Nitpick: Can we move this into a separate `tests/` directory?

For this, there shouldn't be an import issue and I personally prefer to keep the root dir as clean as possible. If there is a reason to keep it here, I don't mind though.

algorithmic_efficiency/workloads/imagenet/imagenet_jax/workload_test.py
```diff
@@ -6,7 +6,7 @@
```
fsschneider, line 6: Nitpick: Can we move this file into a separate `tests/` directory?

For this, there shouldn't be an import issue and I personally prefer to keep the main lib and the tests separate as much as possible. I hope that this would make our code more understandable and easier to navigate.
But if there is a reason to keep it here, I won't object as it really is just personal preference.

algorithmic_efficiency/workloads/imagenet/imagenet_pytorch/workload.py
```diff
@@ -28,7 +30,7 @@ def cycle(iterable):
       iterator = iter(iterable)
 
 
-class ImagenetPytorchWorkload(ImagenetWorkload):
+class ImagenetPytorchWorkload(BaseImagenetWorkload):
```
fsschneider, line 33: Nitpick: I believe it should be `ImagenetWorkload` to be consistent with the other workloads.

However, this is probably just a temporary convention. Once we have multiple workloads all using ImageNet, we need to rethink this anyway. We could use something like `DatasetModelFramework` as our naming convention (e.g. `ImagenetResNetPytorch` in this case, although the `Framework` part might not be necessary.

fsschneider (approved):
Just three tiny nitpicks that are mostly personal preference.
Otherwise it looks good and I can merge.

algorithmic_efficiency/workloads/imagenet/imagenet_pytorch/workload.py
znado: I think that it may be easier to remove the `Framework` from the name, because if we add it then we need to change the workload registry in `submission_runner.py` to have a jax and pytorch class name (it's just easier if they are the same). In that case, we need the `Base` in the name of the parent class, to avoid redefining the class (one of the linter tests complained about it IIRC). I'll change the other workload to have the `Base` convention.

test_version.py
znado: SGTM, done.

algorithmic_efficiency/workloads/imagenet/imagenet_jax/workload_test.py
znado: SGTM, done.

algorithmic_efficiency/workloads/imagenet/imagenet_pytorch/workload.py
fsschneider: Agreed!

PR #64, opened 2022-03-29
Update Rules: Add new loss types and clarify loss != target metric

fsschneider:
Added L1 and CTC loss as possible losses.

Added that the target performance might not be the loss but a different, workload-specific metric such as the accuracy.

Bumped version number.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -1,6 +1,6 @@
 # MLCommons™ Algorithmic Efficiency Benchmark Rules
 
-**Version:** 0.0.3 *(Last updated 22 February 2022)*
+**Version:** 0.0.4 *(Last updated 29 March 2022)*
 
 > **TL;DR** New training algorithms and models can make neural net training faster.
 > We need a rigorous training time benchmark that measures time to result given a fixed hardware configuration and stimulates algorithmic progress. We propose a [Training Algorithm Track](#training-algorithm-track) and a [Model Track](#model-track) in order to help disentangle optimizer improvements and model architecture improvements. This two-track structure lets us enforce a requirement that new optimizers work well on multiple models and that new models aren't highly specific to particular training hacks.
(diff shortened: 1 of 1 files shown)


RULES.md
```diff
@@ -107,7 +107,7 @@ model_fn(
 loss_fn(label_batch, logits_output_batch) -> 1d array of losses per example  # differentiable
 ```
 
-- Unlike in the [Model Track](#model-track), we will specify the loss function name in order to let training algorithms depend on the loss function. It will be one of {**mean squared error**, **cross-entropy**}.
+- Unlike in the [Model Track](#model-track), we will specify the loss function name in order to let training algorithms depend on the loss function. It will be one of {**mean squared error**, **cross-entropy**, **CTC**, or **L1**}.
```
georgedahl, line 110: L1 reconstruction error

RULES.md
```diff
@@ -365,7 +365,7 @@ The target performance on each held-out workload will be defined by using the pe
...
+Submissions will be scored based on their required training time to reach the target performance of each workload. The target performance can be the same as the loss but might also be a different workload-specific metric such as the accuracy or BLEU score. The training time includes compilation times for computation graphs and ops that could happen just-in-time during training; all our benchmarks should be fast enough to compile so as not to dramatically impact overall performance. The overall ranking is then determined by summarizing the performances across all [workloads](#workloads), both public and held-out, using [performance profiles](http://www.argmin.net/2018/03/26/performance-profiles/), as explained below.
```
georgedahl, line 368: "The target performance can ..." -> "The target performance metric can ..."

georgedahl (approved):

RULES.md
georgedahl: Probably should also change "loss" to "loss function"

Issue #62, opened 2022-03-11
MNIST fails to run on main branch


danielsnider:
Recent merges have caused some bugs when trying to run the mnist workload (both jax and pytorch).

## Description

The bugs were introduced sometime between now and this commit (which works):
```
commit 6f8c6bdc8f4055659d7f7e75a512152dbbd35daa (upstream/main)
Merge: cc58b23 489df83
Author: Frank <frank.stefan.schneider@gmail.com>
Date:   Wed Feb 23 11:10:59 2022 +0100

    Merge pull request #57 from fsschneider/rules_update_dependencies

    Add suggested rule changes RE dependencies
```

## Steps to Reproduce

```
python3 submission_runner.py --framework=pytorch --workload=mnist_pytorch --submission_path=baselines/mnist/mnist
_pytorch/submission.py --tuning_search_space=baselines/mnist/tuning_search_space.json
```
or

```
python3 submission_runner.py --framework=jax --workload=mnist_jax --submission_path=baselines/mnist/mnist
_jax/submission.py --tuning_search_space=baselines/mnist/tuning_search_space.json
```

## Source or Possible Fix

I found three bugs, but there are possibly more.

### Bug 1. `README.md` is out of date.

The `algorithmic_efficiency/submission_runner.py` is now located at `submission_runner.py`, but the README.md doesn't reflect this change.

### Bug 2. `random_utils` import issue 

```
$ python3 submission_runner.py --framework=jax --workload=mnist_jax --submission_path=baselines/mnist/mnist_jax/sub
mission.py --tuning_search_space=baselines/mnist/tuning_search_space.json

Traceback (most recent call last):
  File "submission_runner.py", line 317, in <module>
    app.run(main)
  File "/root/.local/lib/python3.8/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "/root/.local/lib/python3.8/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "submission_runner.py", line 302, in main
    workload = _import_workload(
  File "submission_runner.py", line 132, in _import_workload
    workload_module = importlib.import_module(workload_path)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
[...]

PR #65, opened 2022-03-30
(Almost) fix MNIST example

fsschneider:
Addresses the bugs described in issue #62.

This should, however, be seen as a hotfix, not as a proper solution.

We currently don't consistently use the `num_X_examples` property for evaluating the model. I would say, ideally, all models should follow the same protocol during `eval_model`, e.g. evaluate on the train set, validation set, and test set.
This would use the `num_X_examples` property, but currently, we are not doing this for most models.

Similarly, we wanted to provide both the `param_shapes` and the `model_params_types` as properties for submitters, but in most workloads, they are not implemented. Ideally, they would be defined in the framework-agnostic base class of each workload.

Lastly, there is the issue that PR #58 introduced an additional `_mask_batch` argument in a couple of places. This is, for example, prominent in the `submission_runner`, which now expects the `data_selection` (a submission function) to also return a `selected_train_mask_batch` variable:

https://github.com/mlcommons/algorithmic-efficiency/blob/c36557c7d8f2411fc316ea33449d0a93640b1252/submission_runner.py#L174-L182

This currently breaks the MNIST example.

We could fix this, by modifying the `data_selection` function of the MNIST submission 
```diff
-return next(input_queue)
+return next(input_queue) + [None]
```
but I am not convinced this is the most elegant solution (a full fix requires a few additional but similar changes).

@znado perhaps you could chip in, but I believe the introduction of the `_mask_batch` arguments currently is also not in line with the API specification we present in the rules: https://github.com/mlcommons/algorithmic-efficiency/blob/main/RULES.md#specification

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Docker is the easiest way to enable PyTorch/JAX GPU support on Linux since only 
 ### JAX
 
 ```bash
-python3 algorithmic_efficiency/submission_runner.py \
+python3 submission_runner.py \
     --framework=jax \
     --workload=mnist |\
     --submission_path=baselines/mnist/mnist_jax/submission.py \
@@ -138,7 +138,7 @@ python3 algorithmic_efficiency/submission_runner.py \
 ### PyTorch
 
 ```bash
-python3 algorithmic_efficiency/submission_runner.py \
+python3 submission_runner.py \
     --framework=pytorch \
     --workload=mnist \
     --submission_path=baselines/mnist/mnist_pytorch/submission.py \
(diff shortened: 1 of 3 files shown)


znado:
We unfortunately need the mask batch for the GNN workload, so unless we can think of a better way of adding it to the spec I just added it everywhere. For MNIST I implemented the num examples, param types, param shapes, and a nicer solution to the mask batch issue in https://github.com/mlcommons/algorithmic-efficiency/pull/71, so we should go ahead and merge this.

fsschneider:
Closing this as PR #71 provides a nicer solution!

PR #67, opened 2022-04-05
feat(librispeech): add working jax code

ClashLuke:
The training is still going, but it looks pretty good so far.\
As discussed, the Jax version adds significant padding (which isn't there in the PyTorch version), which should cause it to run 2x to 16x slower. At the same time, I had to add masks to BatchNorm, which will cause the performance of this model to deviate slightly from that of the PyTorch model.
There were also many issues with the original Jax submission, so we should probably double-check with the PyTorch version and avoid these in the Conformer implementation.

grafik

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -48,6 +48,10 @@ WORKLOADS = {
   'librispeech_pytorch': {
     'workload_path': 'workloads/librispeech/librispeech_pytorch/workload.py',
     'workload_class_name': 'LibriSpeechWorkload'
+  },
+  'librispeech_jax': {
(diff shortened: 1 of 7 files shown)


ClashLuke:
We now have [preliminary logs](https://github.com/mlcommons/algorithmic-efficiency/files/8459994/baseline.txt) showing the poor stability of our model: cer

ClashLuke:
```
python3 -m pip install --upgrade pip
python3 -m pip install --upgrade "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
python3 -m pip install flax

znado:
cleaning up old PRs, closing

PR #68, opened 2022-04-06
Moving baselines/ to references_submissions/

znado:
Now that we will have actual baseline submissions (the same training algo across all workloads), we should separate out the reference submissions we've been using to test the code. I'm open to suggestions on different names for the dir.

In a follow-up PR I'll make tests that run a train and eval step for each workload.


fsschneider (approved):
Looks good to me! Moving them out of the `baselines/` dir is a great idea.

Two small things:

1. Should we update our `README.md` to reflect these changes? E.g. the `Running a workload` section would need to change the `submission_path` and `tuning_search_space` args to change from `baselines/` to the new dir. As soon as we have a (real) baseline, we might want to change it back again, since running a submission is the more common use case compared to running a "reference submission".
2. I am a bit hesitant to call them "reference submissions" mainly because they aren't proper submissions. I think, under the term "reference submission" I would expect sample submissions of popular algorithms such as Adam, etc. Unfortunately, I don't have a better name for it. We could think about "workload-specific baselines" (calling what is now "baselines" "submission baselines" instead) but this is quite unwieldy. Perhaps something like "reference training" could work?

But I am happy with the PR, feel free to merge.

PR #80, opened 2022-06-12
WIP Jax ViT workload

znado:
This is likely not fully complete, but I want to get a template out there so that the PyTorch implementation has something to go off of.

- renames the imagenet workload and reference submission dirs to imagenet_resnet (and adds an imagenet_vit dir), we will need to clean up dir structure at some point in the future
- forks the [ViT model from init2winit](https://github.com/google/init2winit/blob/master/init2winit/model_lib/vit.py) (the codebase we use internally to run experiments at Google)
- reuses a lot of code from the imagenet resnet workload
- directly copies the imagenet resnet reference submission, which may be different than the one we eventually use to test the correctness of the vit workload
- passes the reference submission test (which uses fake data)


fsschneider:
Superseeded by #82

PR #81, opened 2022-06-13
Fix bugs and implement DDP

runame:
Changes implemented in this PR:

- Implement DDP for MNIST + ImageNet
- Fix MNIST and ImageNet bugs
- Add `data_utils.py` file which collects helper functions related to the input pipelines

There will be a follow-up PR which covers:

- Add DDP to WMT worklod
- Potentially speed-ups for ImageNet and WMT workload
- Maybe improve interface, to enable user to choose number of GPUs to use (right now it simply uses all available GPUs)

### DIFF
--- a/algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
+++ b/algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
@@ -9,7 +9,7 @@ import jax
 import tensorflow as tf
 import tensorflow_datasets as tfds
 
-import algorithmic_efficiency.data_utils as data_utils
+from algorithmic_efficiency import data_utils
 
 IMAGE_SIZE = 224
 RESIZE_SIZE = 256
--- a/algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
+++ b/algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
@@ -9,6 +9,8 @@ import jax
 import tensorflow as tf
 import tensorflow_datasets as tfds
 
+import algorithmic_efficiency.data_utils as data_utils
+
 IMAGE_SIZE = 224
 RESIZE_SIZE = 256
 MEAN_RGB = [0.485 * 255, 0.456 * 255, 0.406 * 255]
@@ -156,16 +158,18 @@ def preprocess_for_train(image_bytes,
   Returns:
     A preprocessed image `Tensor`.
   """
-  crop_rng, flip_rng = tf.random.experimental.stateless_split(rng, 2)
+  # Note (runame): Cannot be done in graph mode, i.e. during ds.map().
+  # Alternative?
+  # crop_rng, flip_rng = tf.random.experimental.stateless_split(rng, 2)
 
   image = _decode_and_random_crop(image_bytes,
-                                  crop_rng,
+                                  rng,
                                   image_size,
                                   aspect_ratio_range,
                                   area_range,
                                   resize_size)
   image = tf.reshape(image, [image_size, image_size, 3])
-  image = tf.image.stateless_random_flip_left_right(image, seed=flip_rng)
+  image = tf.image.stateless_random_flip_left_right(image, seed=rng)
   image = normalize_image(image, mean_rgb, stddev_rgb)
   image = tf.image.convert_image_dtype(image, dtype=dtype)
   return image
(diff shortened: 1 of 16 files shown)


algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
```diff
@@ -156,16 +158,18 @@ def preprocess_for_train(image_bytes,
   Returns:
     A preprocessed image `Tensor`.
   """
-  crop_rng, flip_rng = tf.random.experimental.stateless_split(rng, 2)
+  # Note (runame): Cannot be done in graph mode, i.e. during ds.map().
+  # Alternative?
+  # crop_rng, flip_rng = tf.random.experimental.stateless_split(rng, 2)
```
runame, line 163: @znado This doesn't work in graph mode (see my comment), what do you think is the best alternative?

algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
```diff
@@ -209,22 +213,19 @@ def create_split(split,
                  aspect_ratio_range=(0.75, 4.0 / 3.0),
                  area_range=(0.08, 1.0)):
   """Creates a split from the ImageNet dataset using TensorFlow Datasets."""
+  del num_batches
   if split == 'eval_train':
-    split = 'train'
+    split = 'train[:50000]'
```
znado, line 218: good catch. should we instead in the caller function, pass in split='train[:{num_eval_train_examples}]'?

algorithmic_efficiency/workloads/imagenet/imagenet_pytorch/workload.py
```diff
@@ -19,18 +22,10 @@
...
-# from https://github.com/pytorch/pytorch/issues/23900#issuecomment-518858050
-def cycle(iterable):
-  iterator = iter(iterable)
-  while True:
-    try:
-      images, labels = next(iterator)
-      yield {'inputs': images, 'targets': labels}
-    except StopIteration:
-      iterator = iter(iterable)
+PYTORCH_DDP = 'LOCAL_RANK' in os.environ
```
znado, line 25: nit: should we rename this `USE_PYTORCH_DDP` ?

znado (approved):

algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
runame: Something like "stateless_split iterating over `tf.Tensor` is not allowed in Graph execution" (according to my google search history lol). It's especially weird because [here](https://github.com/mlcommons/algorithmic-efficiency/pull/81/files/c4a07ac76aa311af883886579344581cc4f9394c#diff-60eed9f70147177b1d33b389535916e760ef922b378efac41e537ef089b85c95R225), [`stateless_fold_in()`](https://github.com/tensorflow/tensorflow/blob/v2.9.1/tensorflow/python/ops/stateless_random_ops.py#L204-L242) works just fine and is also called inside of the `ds.map()` call, i.e. is also executed in graph mode. And it is calling the same underlying function. So the issue might be related to the shape argument in `stateless_random_uniform()`, but I haven't looked into it further.

algorithmic_efficiency/workloads/imagenet/imagenet_jax/input_pipeline.py
runame: Makes a lot of sense, already changed it.

algorithmic_efficiency/workloads/mnist/mnist_pytorch/workload.py
```diff
@@ -101,18 +115,31 @@ def build_input_queue(self,
...
-                        global_batch_size: int):
-    return self._build_dataset(data_rng, split, data_dir, global_batch_size)
+                        global_batch_size: int) -> Dict[str, Any]:
+    it = self._build_dataset(data_rng, split, data_dir, global_batch_size)
+    for batch in it:
+      if isinstance(batch, dict):
+        inputs = batch['inputs']
+        targets = batch['targets']
+      else:
+        inputs, targets = batch
```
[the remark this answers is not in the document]
runame: The custom [`cycle`](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/data_utils.py#L27-L39) function returns a dict, but is only used for training here, because for testing `itertools.cycle` is used, which catches the whole dataset in memory, which shouldn't happen during training even if the size of the dataset is no issue because it wil not re-shuffle. But I don't think it is less efficient to just use my custom cycle function also for the test set, so we can get rid of the if-statement (and it is consistent with e.g. the ImageNet workload).

algorithmic_efficiency/workloads/imagenet/imagenet_pytorch/workload.py
fsschneider: `USE_PYTORCH_DDP` sounds good to me

setup.cfg
```diff
@@ -142,7 +142,7 @@ profile=google
 # pylint configuration
 [pylint.MASTER]
 persistent=no  # Pickle collected data for later comparisons.
-cache-size=500  # Set the cache size for astng objects.
+#cache-size=500  # Set the cache size for astng objects.
```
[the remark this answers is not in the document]
runame: Yes, these options are not supported anymore in pylint>=2.14.0. Sounds reasonable to switch to the pylintrc from the Google Style Guide.

PR #85, opened 2022-06-21
Add DDP to WMT + faster ImageNet data loading

runame:
Follow-up PR to #81.

This PR 
- adds DDP to WMT PyTorch workload,
- fixes bug in WMT Jax workload,
- improves the data loading for ImageNet PyTorch workload,
- adds instructions for running DDP to README.

One question specifically for @mikerabbat: I have implemented the distributed sampling for the TF dataset by simply manually sharding each global batch across devices, see [this wrapper](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/data_utils.py#L219-L244) which assumes [this function](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/data_utils.py#L10-L27) has been mapped on the dataset before. I think you mentioned at some point that there is an issue with this, but I'm not sure what it was.

And one question @znado might be able to help with: when running the DDP WMT PyTorch workload with 8xV100s instead of 4xV100s, I'm getting this error:
```bash
Check failed: ret == 0 (11 vs. 0)Thread tf_pjrt_thread_pool creation via pthread_create() failed.
```
I think this is caused by the fact that the TF input pipeline is created for each process separately, i.e. as many times as there are GPUs. The error seems to indicate that this exhausts the resources, see [this explanation](https://github.com/tensorflow/tensorflow/issues/41532#issuecomment-759075803). I tried setting `AUTOTUNE = None` ([here](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/workloads/wmt/input_pipeline.py#L12)) and increasing the number of threads which are allowed per process (using `torch.set_num_threads(N)` with `N` up to 8; the default ist `N=1`), but both didn't help.


znado:
I would have hoped that setting `AUTOTUNE = None` would have fixed that out of threads issue. Beyond the number of threads allowed per process, do you know if the host machines you're running on have a low number of max threads allowed (I could see that being the case on some clusters to avoid accidental fork bombs)? I'm not sure the most relevant place to check, but I know there are some places you can check at the Linux/system level (you may already have, just checking!).

As for sharding the global batch across GPUs like you do, would it instead be more efficient to run each copy of the input pipeline with the per-GPU batch size? Then you don't need to slice the batch (like you do [here](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/data_utils.py#L241)), you would just need to add a toggleable flag to the input pipeline to not [shard](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/data_utils.py#L10-L27) the batches. Maybe this would help avoid memory issues (although it looks like the resource issue you're having is with threading, not memory?) To make sure you get a unique batch on each process, you can (hopefully?) fold the process rank into the seed, for example you could add this [here](https://github.com/mlcommons/algorithmic-efficiency/blob/987a47ea873f9777cd58b6d6fffa006adce4b158/algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py#L235):
```
data_rng = jax.random.fold_in(data_rng, torch.distributed.get_rank())
np_iter = super().build_input_queue(data_rng, ...)
```

algorithmic_efficiency/workloads/wmt/input_pipeline.py
```diff
@@ -211,7 +211,7 @@ def preprocess_wmt_data(dataset: tf.data.Dataset,
                         train: bool,
                         shuffle_buffer_size: int = 1024,
                         max_length: int = 512,
-                        per_device_batch_size: int = 256):
+                        batch_size: int = 256):
```
znado, line 214: nit: can we name this global_batch_size just so we are clear in the future?

algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py
```diff
@@ -192,18 +186,10 @@ def translate_and_calculate_bleu(self,
...
     sources, references, predictions = [], [], []
     for _ in range(num_batches):
       pred_batch = next(ds_iter)
-      # Handle final odd-sized batch by padding instead of dropping it.
-      cur_pred_batch_size = pred_batch['inputs'].shape[0]
-      if cur_pred_batch_size % n_devices:
-        padded_size = int(np.ceil(cur_pred_batch_size / n_devices) * n_devices)
-        pred_batch = jax.tree_map(
-            functools.partial(_pad_examples, padded_size),  # pylint: disable=cell-var-from-loop
-            pred_batch)
```
znado, line 206: why do we need to delete this? we have similar logic for pytorch, right?

znado (approved):
just one question, otherwise this is fantastic, thanks so much for all these fixes!! it's awesome to get the pytorch pipelines sped up

algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py
runame: Because the batch is already [sharded](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/workloads/wmt/input_pipeline.py#L289) (before by calling `ds = ds.batch(...)`), hence there is no need for padding anymore. For PyTorch without DDP, the batch is not sharded at all, but I just realized that in fact it is first sharded and then [reshaped](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py#L259-L260) again, so we can also delete the code block for PyTorch.

Note, that the current way of sharding the batches [assumes](https://github.com/runame/algorithmic-efficiency/blob/pytorch-speedups/algorithmic_efficiency/data_utils.py#L22-L25) that the remainder batch, which is not dropped for evaluation, is evenly divisble by the number of GPUs. This affects all workloads which use this sharding function, e.g. ImageNet and WMT. This might obviously lead to problems, depending on the batch size used in the submissions. One easy solution would be to hard-code the evaluation batch size or provide an error which suggests the closest batch size which works well to the user. For the part where the batch size is more important, i.e. training, this issue does not appear anyway, because the remainder batch is dropped.

runame: Maybe a good solution is to provide the option to choose the training and eval batch size separately and provide a working default for evaluation. If the user really wants to change the evaluation batch size, we can raise an error which suggests the closest working one. This also adds more flexibility for the user, since they might want to use a larger batch size for evaluation than for training.

znado: Yeah having different batch sizes for train and eval is super useful in general, we should definitely support it. I'll add a TODO for myself to look into it (unless you want to soon!)

znado:
Yeah that's a good point regarding possibly repeated examples. A solution we've used to that is sharding the input files across processes, so you guarantee you have different examples per process (and then you don't need to mess with the RNGs).

PR #93, opened 2022-07-13
Jax Criteo input pipeline

znado:
It is unclear if this includes all the preprocessing we need, but making jax input pipeline runs with external data.

I only implemented preprocessing as log(1+x) for dense features, and that the hex strings are parsed into int64, then modded with the vocab size and cast to int32.

- fixing setup.cfg
- fixing readme paths


algorithmic_efficiency/workloads/criteo1tb/input_pipeline.py
```diff
@@ -39,13 +39,15 @@ def _parse_example_fn(example):
...
+      int_features.append(tf.math.log(fields[idx + num_labels] + 1))
     int_features = tf.stack(int_features, axis=1)
 
     cat_features = []
     for idx in range(len(vocab_sizes)):
       cat_features.append(
-          tf.cast(fields[idx + num_dense + num_labels], dtype=tf.int32))
+          tf.io.decode_raw(fields[idx + num_dense + num_labels], tf.int64)[:, 0]
+          % vocab_sizes[idx])
+    print(cat_features)
```
runame, line 50: Do you want to keep the `print` statement here or was this just for debugging?

runame (approved):
I can't really tell if the preprocessing is sufficient/correct, but the rest LGTM.

algorithmic_efficiency/workloads/criteo1tb/input_pipeline.py
znado: ah thanks for catching that, removed.

PR #98, opened 2022-07-21
Minor bug fixes for cifar_jax

pomonam:
This PR fixes minor bugs in the `cifar_jax` workload.

- `train_eval` split should not drop the last batch.
- `ds.repeat()` only when `repeat_final_dataset` is set to True.
- Bug fix when performing multi-GPU evaluation in `_eval_model_on_split`.
- General refactor following `imagenet_resnet` for consistency.


algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
```diff
@@ -19,6 +21,29 @@
 
 class CifarWorkload(BaseCifarWorkload):
 
+  def __init__(self):
+    super().__init__()
+    self._param_shapes = None
+    self._param_types = None
+    self.epoch_metrics = []
+    self._eval_iters = {}
```
runame, line 29: All of these attributes besides `self.epoch_metrics` are already defined in `BaseCifarWorkload`. Also, I have noticed that the `__init__()` method of the PyTorch version of the workload can be deleted.

algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
```diff
@@ -140,27 +180,51 @@ def loss_fn(self, label_batch: spec.Tensor,
...
+      eval_iter = self.build_input_queue(
+          data_rng,
+          split=split,
+          global_batch_size=global_batch_size,
+          data_dir=data_dir,
+          cache=True,
+          repeat_final_dataset=True,
+          num_batches=num_batches)
+      # Note that this stores the entire eval dataset in memory.
+      self._eval_iters[split] = itertools.cycle(eval_iter)
```
runame, line 217: This is not really necessary, since `repeat_final_dataset=True`, right? Unless we actually want to keep the eval sets in memory; however, this is not the case for the PyTorch version. Also, won't the memory consumption not increase with every evaluation, since the iterator never hits `StopIteration`, i.e. the data which is stored in memory is actually not used for the next evaluation?

runame (approved):
Besides the two points in the comments, LGTM!

algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
pomonam: Thanks for the check -- fixed!

algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
pomonam: That is correct -- I removed this line.

runame:
Thanks for the fixes! The pylint and isort checks broke, could you also fix them?

pomonam:
@runame It seems like isort and pylint are failing on the main branch. For example, below is the output of isort. 

```
ERROR: /home/runner/work/algorithmic-efficiency/algorithmic-efficiency/algorithmic_efficiency/workloads/librispeech/prepare_data.py Imports are incorrectly sorted and/or formatted.
```

runame:
> It seems like isort and pylint are failing on the main branch. For example, below is the output of isort.

Ah sorry, you are right. I'll create a new PR soon anyway and will fix this.

PR #99, opened 2022-07-21
fixes to existing librispeech pytorch workload and a .gitignore to ig…

sourabh2k15:
1) added .gitignore to remove generated librispeech from commits 
2) fixes to librispeech pytorch workload as per reference implementation here : https://github.com/lsari/libri100_conformer_ctc/blob/main/test_wer.py 

currently this pytorch code is broken cause the exisitng ctcdecode package is not possible to install

### DIFF
--- a/algorithmic_efficiency/workloads/librispeech/.gitignore
+++ b/algorithmic_efficiency/workloads/librispeech/.gitignore
@@ -1,0 +1,5 @@
+data_dir/
+work_dir/
+*.flac
+*.npy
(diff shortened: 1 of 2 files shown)


sourabh2k15:
I have filled the form that CLA bot points above to and I have also signed the internal CLA for googlers to contribute in open source repos. 

it works for other open source github repos by google so it seems there's a MLCommons specific one I might not have signed, trying to find that CLA

PR #101, opened 2022-07-26
Fix DDP issues

runame:
This PR
- switches to a single process input pipeline for the PyTorch workloads, which use a tf input pipeline (WMT + OGBG). This avoids tf creating too many threads. While the code is a bit more convoluted, the change not only fixes the issue, but according to my preliminary tests also seems to either not affect speed or even improve it.
- It fixes a DDP time-out issue related to entering evaluation at different steps,
- and fixes some other minor stuff (requirements, warnings, ...).

TODO:
- [ ] Add a similar input pipeline to the PyTorch ImageNet workload, to avoid using two separate datasets.


README.md
```diff
@@ -58,6 +58,12 @@
    pip3 install -e '.[full]'
    ```
 
+   For the PyTorch version of the ogbg workload, you might have to install
```
znado, line 61: nit: is it actually required to install? if so can we change "you might have to" to "you need to"

algorithmic_efficiency/workloads/cifar/cifar_pytorch/workload.py
```diff
@@ -77,18 +77,19 @@ def _build_dataset(self,
...
-                train=split in ['train','eval_train','validation'],
-                download=True,
-                transform=train_transform_config if "train" in split \
-                                            else eval_transform_config)
+    dataset = CIFAR10(
+        root=data_dir,
+        train=split in ['train', 'eval_train', 'validation'],
+        download=True,
+        transform=train_transform_config
+        if 'train' in split else eval_transform_config)
```
znado, line 85: nit: can you pull this out into a new line above? I try to avoid multi-line ternary operators

algorithmic_efficiency/workloads/wmt/input_pipeline.py
```diff
@@ -7,9 +7,11 @@
 import tensorflow_datasets as tfds
 
 from algorithmic_efficiency import data_utils
+from algorithmic_efficiency.pytorch_utils import pytorch_setup
 from algorithmic_efficiency.workloads.wmt import tokenizer
 
-AUTOTUNE = tf.data.AUTOTUNE
+RANK = pytorch_setup()[1]
+AUTOTUNE = tf.data.AUTOTUNE if RANK == 0 else None
```
znado, line 14: is this saying, set to None if on the first process? can you add a quick comment stating that?

algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
```diff
@@ -226,26 +216,66 @@ def build_input_queue(self,
...
+                                              dtype=torch.int32,
+                                              device=DEVICE)
+          dist.broadcast(per_device_batch_size, src=0)
+        tensor = torch.empty((n_inputs, N_GPUS, per_device_batch_size, 256),
+                             dtype=torch.int64,
+                             device=DEVICE)
+        dist.broadcast(tensor, src=0)
+        # Note that the order of the keys is important.
+        # For split == 'train'.
+        if n_inputs == 6:
```
znado, line 263: nit: can you replace this with `if split == 'train': ... else` to be a bit clearer?

znado (approved):
Looks good (and complicated) to me! Thanks so much for untangling the gross cross-framework threading issues, this is going to be critical to getting the benchmark launched

algorithmic_efficiency/workloads/wmt/input_pipeline.py
runame: It should only use `tf.data.AUTOTUNE` in the first process (rank 0), to avoid multithreading in all other processes. If PyTorch DDP is not used, e.g. when Jax is being used, `RANK` will be `0`, so `AUTOTUNE` will also be set correctly. I have added a clarifying comment.

PR #102, opened 2022-07-28
Clean up & Fix bugs in imagenet_vit

pomonam:
This PR does the following:
- Clean-up code in `imagenet_vit` workload.
- Set hyperparameter configuration according to https://github.com/google/init2winit/blob/master/init2winit/model_lib/vit.py (e.g., use Adam optimizer with base LR 1e-3).
- Fix bug in `imagenet_pytorch` workload when `rep_size=True`.

I performed some profiling on top of this PR on 4 GPU machines (with a 10% validation target) and confirmed that the speed difference between PyTorch and Jax is within 10%.

### DIFF
--- a/algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/workload.py
@@ -1,5 +1,4 @@
 """ImageNet workload implemented in Jax."""
-import math
 from typing import Dict, Tuple
 
 from flax import jax_utils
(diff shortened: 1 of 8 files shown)


pomonam:
It seems like the linting is failing for the same reason as https://github.com/mlcommons/algorithmic-efficiency/pull/98#issuecomment-1194540378 and PR #101 should be able to fix this.

algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/workload.py
```diff
@@ -46,6 +47,40 @@ def model_fn(
...
+  def _eval_model_on_split(self,
+                           split: str,
+                           num_examples: int,
+                           global_batch_size: int,
+                           params: spec.ParameterContainer,
+                           model_state: spec.ModelAuxiliaryState,
+                           rng: spec.RandomState,
+                           data_dir: str):
+    data_rng, model_rng = jax.random.split(rng, 2)
+    num_batches = int(math.ceil(num_examples / global_batch_size))
```
runame, line 62: If I'm not missing anything, the only difference to the same method of the parent class ` ImagenetResNetWorkload` is this line:
```python
# Sync batch statistics across replicas before evaluating.
model_state = self.sync_batch_stats(model_state)
```
Maybe we can avoid redundancy by doing something like in `ImagenetResNetWorkload`:
```python
if model_state is not None:
  # Sync batch statistics across replicas before evaluating.
  model_state = self.sync_batch_stats(model_state)
```
If `model_state` is not None anyway in the ViT workload, we can set it to None before inheriting from the parent class, i.e.
```python
model_state = None
return super()._eval_model_on_split(
    split,
    num_examples,
    global_batch_size,
    params,
    model_state,
    rng,
    data_dir)
```

runame (approved):
LGTM!

algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/workload.py
pomonam: Thanks for the review! Yes, you are correct -- I made the modification accordingly.

znado (approved):
Now that we have merged PR #101, I think if you sync and push again this should be good to go?

pomonam:
Yes, it is ready to be merged now.

PR #109, opened 2022-08-02
Librispeech Conformer JAX workload

sourabh2k15:


sourabh2k15:
this is still WIP as I'm trying to run it e2e on VM but the code is mostly here

PR #110, opened 2022-08-04
Fix our linting tests.

fsschneider:
Just small changes that fix our linting issues and pass the tests.


runame (approved):
Fixed this in #101, but doesn't hurt to merge this in a separate PR now.

PR #117, opened 2022-08-15
Rules update

fsschneider:
Suggests an update to the rules according to the discussions in the working group meetings.
Specifically, the following topics have been changed:

- [x] Awards and prize money
- [x] Setting the evaluation frequency, e.g. `eval_period`
- [x] Description of the qualification set
- [x] Performance check on the test set
- [x] Generation of held-out workloads
- [x] Target-setting procedure on both the public and the held-out workloads
- [x] Scoring procedure
- [x] Longer runtime budget for self-tuning
- [x] ImageNet will use a test set now (remove prior statement)

---

Moved into next PR:

- [ ] Multiple submissions
- [ ] Explanation for using the validation set for tuning, but test set for scoring (use test set for self-tuning ruleset?)
- [ ] Tuning on different hardware
  - [ ] Submitters need to use fastest-to-validation target even when self-reporting
- [ ] Specify benchmarking hardware

There were also some additional formatting changes, e.g. fixing math expressions, code layout, etc.


RULES.md
```diff
@@ -38,13 +40,13 @@ The goal of the Training Algorithm Track is to reach the same results faster ("t
...
 
 The intention is that a training algorithm submission will be broadly applicable and useful without customization to the specific [workload](#workloads) (model, dataset, loss function). We want to discourage detecting the particular workload and doing something highly specific that isn't generally useful. In order to further discourage submissions that overfit to the particular [public benchmark workloads](#public-workloads), submissions must also perform well on one or more [held-out workloads](#held-out-workloads) released after the submission deadline.
```
mikerabbat, line 45: > submissions must also perform well on one or more [held-out workloads](#held-out-workloads) released after the submission deadline.

performing "well on one or more workload" is vague. Perhaps "submissions will also be evaluated on [held-out workloads]..." instead?

RULES.md
```diff
@@ -38,13 +40,13 @@ The goal of the Training Algorithm Track is to reach the same results faster ("t
...
... submission is a piece of code with the same high-level structure as a reference implementation that can train all benchmark workloads on the [competition hardware](#competition-hardware) (defined in the [Scoring](#scoring) section but ultimately in the call for submissions). The validation set performance will be checked regularly during training (see the [Evaluation during training](#evaluation-during-training) section) and training halts when a workload-specific [target error](#defining-target-performance) has been reached. For each workload, the training time to reach this (validation set) target error will be used as an input to the [scoring process](#scoring) for the submission. Additionally, the test set performance will be probed using the final model parameters to confirm that it also reaches a slightly more generous target performance on this unseen data (see the [scoring](#scoring)] section for details). Submissions using [external tuning](#external-tuning-ruleset) will be tuned independently for each workload using a single workload-agnostic search space for their specified hyperparameters. Submissions under either tuning ruleset may always self-tune while on the clock.
```
mikerabbat, line 49: > A valid submission

is a piece of code **that defines all of the submission functions**

> (defined in the [Scoring](#scoring) section but ultimately in the call for submissions)

Can we just leave it as either "defined in the scoring section" or "to be specified in the call for submissions"? I guess the latter is preferable if we expect the competition hardware to change from call to call.

Nit: it might be more consistent to drop "competition" and use "benchmark" throughout the rules. (Is the competition the same as the benchmark?)

RULES.md
```diff
@@ -74,7 +76,7 @@ def build_input_queue(
 ###### Model initialization
```
mikerabbat, line 76: The line just above this is describing the `build_input_queue` function, and it mentions

> ... and for PyTorch this should return tensors of shape `(global_batch_size, ...)`.

Is this still correct, or should it instead be `(per_device_batch_size, ...)`, which is the usual way that data loading is handled with multi-gpu training in PyTorch (with a DistributedSampler, so that each device gets different samples).

cc @runame

RULES.md
fsschneider: @mikerabbat Do I understand it correctly that you would like to change the sentence:

"A valid submission is a piece of code with the same high-level structure as a reference implementation that can train all benchmark workloads on the [competition hardware](#competition-hardware) (defined in the [Scoring](#scoring) section but ultimately in the call for submissions)."

to something like:

"A valid submission is a piece of code that defines all of the submission functions and is able to train all benchmark workloads on the [competition hardware](#competition-hardware) (defined in the [Scoring](#scoring) section)."?

I would expect us to update this rules document for each iteration of the benchmark and so we could update the competition hardware (or should I say: benchmark hardware) from call to call in this doc.

RULES.md
```diff
@@ -317,90 +319,133 @@ Valid submissions must rely on new algorithmic or mathematical ideas and should
...
 
-- Suggestion: *S=5*, *O=20*.
-- The number of trials that will be performed is known to the submissions.
-- Run on 20 machines in parallel, report the minimum time per study.
-- Submissions should work with a simple random search, within the provided search space.
-- To estimate study variance and to rule out lucky studies, *S* studies will be run.
-- Submissions are always free to perform additional self-tuning while being timed.
+In each study, the fastest training time across the $O=20$ settings will be taken into account and the median of these $5$ per-study training times will be the final training time for the submission on this workload (see [Scoring submissions](#scoring) section). Runs that do not reach the target performance of the evaluation metric have an infinite time. Submissions are always free to perform additional self-tuning while being timed.
```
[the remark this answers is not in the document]
fsschneider: We discussed this in the meeting on June 23rd (see Minutes: https://docs.google.com/document/d/1eMSebuRvSp4vVdb26lJgEv6Lj7mMz5-KOdBjgYfRpck/edit#heading=h.6nd8neqte1l9)

I am happy to quickly discuss this again. I can put it on the agenda for next week. Could you perhaps in this meeting explain your alternative proposal? If I understand you correctly, you would essentially produce 5 performance profile curves for a submission, right? But you would only use the min across all 5 studies for the competition score?

RULES.md
fsschneider: I consistently use "benchmark" now instead of "competition", including calling the the "benchmarking hardware" and the 
(scalar) "benchmarking score". Thanks for the suggestion!

RULES.md
runame: It actually depends on whether DP or DDP is used. In the former case, the current statement `(global_batch_size, ...)` is correct; in the latter, `(per_device_batch_size, ...)` is correct.

RULES.md
mikerabbat: > to something like:
> 
> "A valid submission is a piece of code that defines all of the submission functions and is able to train all benchmark workloads on the [competition hardware](https://github.com/mlcommons/algorithmic-efficiency/pull/117#competition-hardware) (defined in the [Scoring](https://github.com/mlcommons/algorithmic-efficiency/pull/117#scoring) section)."?

Yes, exactly!

RULES.md
mikerabbat: We should really only be using DDP

RULES.md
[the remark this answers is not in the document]
mikerabbat: In the current rules, the performance profile for a submission is based on its performance on W workloads. (W=8 if we only consider public workloads)
The score on each workload is relative to the best performing submission on that workload, i.e., $\min_{s} t_{w,s}$ where $t_{w,s}$ is the time it takes submission $s$ to reach the targets on workload $w$.

Currently we get $t_{w,s}$ for each submission by taking the median over 5 seeds. Let's call $t_{w,s,i}$ the time to reach targets of submission $s$ on workload $w$ for the $i$ th seed.

Another way we could capture the "performance distribution" of a submission would be to use all of the individual times. So instead of first computing medians and then using W numbers to compute the performance profile of a submission, we would directly use all 5W numbers to compute the profile. The best time for the workload (reference for computing the performance profile) would need to be $\min_s \min_i t_{w,s,i}$ so that all curves still have support $[1, \infty)$.

Why even consider this? It relates to the question you asked at that June 23 meeting. Suppose that a submission gets fast times on 3 out of their 5 seeds, and the other two are very slow. Currently they would only be scored based on the fast time. In the more complex alternative setup, their curve would be impacted by the two slow runs. Similarly, if a submission is fast 2 out of 5 seeds, and the other three are slow, the submission's score in the current rules will be the slower one.

Using all 5 times would reward approaches that have low variance/variability and penalize approaches that have higher variance (but sometimes do exceptionally well).

RULES.md
fsschneider: As discussed in the meeting, I will create a second PR with all remaining rules changes where we can address this.
I can change it to `per_device_batch_size`. Should I also add something about DDP?

runame: Maybe just add something like "(assuming DDP is used)". I agree that in practice only DDP should be used, but it is useful for debugging to also support DP in the code.

fsschneider: Done.

RULES.md
```diff
@@ -317,90 +319,131 @@ Valid submissions must rely on new algorithmic or mathematical ideas and should
...
+#### Awards and prize money
 
-To have a simpler to interpret number for press releases, we will also release (in addition to the raw  $t(p,s) $ values) the geometric mean of  $\frac{t_{p, \text{ref}}}{t_{p,s}} $.
+An awards committee will award a prize for the "*Best Performance*" in each ruleset as well as a "*Jury Award*". The prize for the best-performing submission will be awarded based on the [benchmark score](#benchmark-score-using-performance-profiles) on the full benchmark including [held-out workloads](#held-out-workloads). The "*Jury Award*" will favor more out-of-the-box ideas that show great potential, even though the method may not be of practical value with the current landscape of models, software, etc.
 
-- Once we fix a reference we can rerun the reference on the new set of problems for each iteration of the contest (using the new competition hardware), and then report our year over year progress as a community in speeding up training
+The prize money for "*Best Performance*" in a ruleset is 20.000$ each. The winner of the "*Jury Award*" will be awarded 10.000$. We reserve the right to split the prize money and distribute it among multiple submissions.
```
georgedahl, line 444: Since the prizes are denominated in US$ let's format this as
$20,000 and $10,000.

georgedahl (approved):

RULES.md
fsschneider: 20$ would be a bit cheaper though 😁

Issue #105, opened 2022-07-28
WMT Jax workload does not converge and hangs


runame:
The WMT Jax workload does not converge in terms of train/validation loss and hangs mid-training (the job keeps running, but no logging messages are printed anymore).

## Description

See above. Notably, the workload worked correctly on January 11, 2022.

## Steps to Reproduce

Run WMT Jax workload using the standard instructions.

## Source or Possible Fix

No idea regarding the source, but the bug had to be introduced in one of the refactors, since the workload worked correctly before.

runame:
Update: The cause of the lack of convergence is a typo in `model_fn()`, which leads to uncomplete inputs to the forward pass; see [this commit](https://github.com/runame/algorithmic-efficiency/commit/86bd8673f85c64f02b13dc4138e90c6b2200ee55) for the fix.

However, the training process still gets stuck at seemingly random steps. I verified that it does not hang during `data_selection()`, but in `update_params()`.

runame:
I think the hanging issue could be caused by [this](https://github.com/google/jax/issues/10969). Updating jax and jaxlib to >= 0.3.14 should then solve the issue. However, after the update I get an error:
```
jax._src.traceback_util.UnfilteredStackTrace: jaxlib.xla_extension.XlaRuntimeError: INTERNAL: external/org_tensorflow/tensorflow/compiler/xla/service/gpu/nccl_utils.cc:245: NCCL operation ncclCommInitRank(comm.get(), nranks, id, rank) failed: unhandled cuda error: while running replica 0 and partition 0 of a replicated computation (other replicas may have failed as well).
```
[Here](https://github.com/google/jax/issues/10461) it is suggested that this might be caused by some mismatch of CUDA versions, but I couldn't identify anything like this in my setup. Would be interesting to know if you can reproduce this error after updating to the most recent jax and jaxlib versions or if the update fixes the hanging issue, @priyakasimbeg.

runame:
Small update: as discussed in the last PyTorch engineering meeting I have verified that the workload does not hang on a single GPU.

priyakasimbeg:
Ran this on updated main branch, on the VM with 8 GPUs, with Jax upgraded to and it still hangs :(.

JAX versions
```
jax==0.3.4                                                                                                                                                                         
jaxlib==0.3.2      
```
[...]

PR #123, opened 2022-08-23
Fix Jax WMT workload

runame:
I wanted to wait until the hanging issue (#105) is fixed, but since it might just require a more recent jax version, I think this fix here can already be merged. See #105 for an explanation.

Besides the fix, I have also refactored the reference submission to use `workload.loss_fn()`.

### DIFF
--- a/reference_submissions/wmt/wmt_jax/submission.py
+++ b/reference_submissions/wmt/wmt_jax/submission.py
@@ -4,8 +4,6 @@ import functools
 from typing import Dict, Iterator, List, Tuple
 
 from flax import jax_utils
-from flax import linen as nn
-from flax.training import common_utils
 import jax
 import jax.numpy as jnp
 import optax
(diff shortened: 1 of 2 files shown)


znado (approved):

znado:
LGTM assuming we aren't tuning label smoothing

reference_submissions/wmt/wmt_jax/submission.py
```diff
@@ -98,49 +96,32 @@ def init_optimizer_state(workload: spec.Workload,
 
 @functools.partial(
     jax.pmap,
-    in_axes=(None, None, 0, 0, 0, None, 0),
```
znado, line 101: lets revert all of the changes to this file, because we will need them later when we tune label smoothing in the v2 of the target setting runs (if we ever want to remove this code in the future, we can always git cherry-pick this commit).

znado (changes requested):

reference_submissions/wmt/wmt_jax/submission.py
runame: Maybe it makes sense to just add an additional `label_smoothing` argument to `workload.loss_fn()` if we want to tune label smoothing for all workloads anyway. Alternatively, we would have to allow one-hot labels as input to `workload.loss_fn()`; currently it says [Dense (not one-hot) labels](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/spec.py#L225).

 In any case the submissions should use `workload.loss_fn()`, since it is part of the definition of a workload, i.e. if a user is not using it, their submission could be considered invalid.

znado: sgtm. I can add label_smoothing and dropout to the workload API in a follow-up PR as we discussed

znado (approved):

reference_submissions/wmt/wmt_jax/submission.py
runame: I wanted to suggest the same, I can add them in the target-setting run PR, since that is where they are needed anyway.

PR #125, opened 2022-09-01
Add `label_smoothing` argument to `loss_fn`

runame:
This PR adds a `label_smoothing` argument to `workload.loss_fn`.

I have also deleted `_conform_weights_to_targets` in `algorithmic_efficiency/workloads/criteo1tb/criteo1tb_jax/metrics.py`, since it is not used anywhere.

**Question:**
~~Until now we have used a different implementation of label smoothing for the WMT workload than the standard variation implemented in init2winit, optax, and PyTorch (see this [docstring](https://github.com/google/init2winit/blob/master/init2winit/model_lib/model_utils.py#L158-L169) for an explanation of the difference). Should we stick to the different variation for that workload or switch to the one used for all other workloads?~~

### DIFF
--- a/target_setting_runs/imagenet_resnet/jax_submission.py
+++ b/target_setting_runs/imagenet_resnet/jax_submission.py
@@ -27,7 +27,7 @@ def get_batch_size(workload_name):
     jax.pmap,
     axis_name='batch',
     in_axes=(None, None, 0, 0, 0, 0, 0, None),
-    static_broadcasted_argnums=(0, 1, 7))
+    static_broadcasted_argnums=(0, 1))
 def pmapped_train_step(workload,
                        opt_update_fn,
                        model_state,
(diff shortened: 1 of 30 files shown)


target_setting_runs/imagenet_resnet/jax_submission.py
```diff
@@ -26,15 +26,16 @@ def get_batch_size(workload_name):
 @functools.partial(
     jax.pmap,
     axis_name='batch',
-    in_axes=(None, None, 0, 0, 0, 0, 0),
-    static_broadcasted_argnums=(0, 1))
+    in_axes=(None, None, 0, 0, 0, 0, 0, None),
+    static_broadcasted_argnums=(0, 1, 7))
```
znado, line 30: (same with the other submission.py files) do we need to specify label_smoothing as a static arg? I doubt it matters, but in the rare case it would ever change during training it would trigger a jax recompilation

znado (approved):
Should we also update the other target setting run submissions? We can always do it in a future PR if needed, not necessary for this one.

target_setting_runs/imagenet_resnet/jax_submission.py
runame: I also briefly considered that but at least for the target-setting runs label smoothing is never changing during training or is it? And people can obviously do whatever they want in their own submissions.

runame: Ok, I will just not specify it as static, probably doesn't hurt anyway.

runame:
> Should we also update the other target setting run submissions? We can always do it in a future PR if needed, not necessary for this one.

I only updated the ones where the table lists that they use label smoothing, but you are right that we will want to tune it for all workloads and ideally also in this repo, so I have added it to all target-setting runs.

PR #126, opened 2022-09-02
loosening py req to 3.7

znado:
3.7 is usually the default version that comes with a lot of the ubuntu/debian images, so this is easier than having to upgrade python to 3.9 every time


runame (approved):
Should be fine, if it breaks something we can obviously still switch back later on.

PR #127, opened 2022-09-07
Dockerfile updates

znado:
Updating `Dockerfile` with latest commands to automate GCP VM setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,14 +6,6 @@ FROM nvidia/cuda:11.5.1-cudnn8-devel-ubuntu20.04
 
 RUN apt-get update
 
-# Needed before python3.9 ?
-RUN DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install tzdata
-
-# Install python 3.9.
-# RUN apt-get install -y software-properties-common && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,12 @@
-FROM nvidia/cuda:11.1.1-cudnn8-devel-ubuntu20.04
+# MLCommons Training Algorithms Dockerfile
+#
+# docker build https://github.com/mlcommons/algorithmic-efficiency.git
 
-RUN apt-get update && \
-    apt-get install -y bash \
+FROM nvidia/cuda:11.5.1-cudnn8-devel-ubuntu20.04
```
runame, line 5: We might need to use CUDA 11.6 or 11.7, because it seems like the most recent versions of Jax do not support 11.5 anymore due to their use of NCCL 2.12.12. See issue #105 for more details on the relevance of updating the Jax version and the potential compatibility issue.

runame (approved):
Besides the potential CUDA issue, LGTM.

Dockerfile
znado: great catch, updated to the 11.7 version of the same base image

PR #129, opened 2022-09-08
Fix bug in cache for autoregressive decoding

### LINKED ISSUE #106: Bug in autoregressive decoding/BLEU score
Bug in cache for autoregressive decoding and the distributed calculation of the BLEU score in the PyTorch WMT workload.

## Description

The loss and accurcay is converging as in the Jax version of the workload, but the BLEU score is different. I suspect this is due to a bug in the implementation of the cache for autoregressive decoding. Moreover, I think the current way of calculating the BLEU score might be incorrect when using DDP.

## Steps to Reproduce

Run PyTorch WMT workload via the standard commands.

## Source or Possible Fix

See description above. If somebody wants to tackle this before I get to it, please get in touch for more details.

runame:
Fixes #106 + some cosmetic changes.

@znado Please review #125 first, since this branch is based on it.

### DIFF
--- a/algorithmic_efficiency/spec.py
+++ b/algorithmic_efficiency/spec.py
@@ -224,7 +224,8 @@ class Workload(metaclass=abc.ABCMeta):
       self,
       label_batch: Tensor,  # Dense (not one-hot) labels.
       logits_batch: Tensor,
-      mask_batch: Optional[Tensor] = None) -> Tensor:  # differentiable
+      mask_batch: Optional[Tensor] = None,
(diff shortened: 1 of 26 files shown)


znado (approved):
commits on top of https://github.com/mlcommons/algorithmic-efficiency/pull/125 LGTM!

runame:
I have ran the fixed workload on 8xV100s and can confirm that the BLEU score now also matches the one of the Jax workload.

PR #130, opened 2022-09-12
fixing eval methods post librispeech workload check-in

sourabh2k15:


znado (approved):
lgtm, we'll refactor the tb and other logging in a follow-up

PR #134, opened 2022-09-14
small fixes to get reference submission tests to work on GCP VM

znado:
- README py3.9 -> py3
- setup.cfg changes
- small ref test fix
- wandb import optional

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -24,7 +24,11 @@ from absl import logging
 import tensorflow as tf
 import torch
 import torch.distributed as dist
-import wandb
+try:
+  import wandb
+except ModuleNotFoundError:
+  logging.exception('Unable to import wandb.')
+  wandb = None
 
 from algorithmic_efficiency import halton
 from algorithmic_efficiency import random_utils as prng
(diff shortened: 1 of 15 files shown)


submission_runner.py
```diff
@@ -206,18 +211,17 @@ def train_once(
...
+                                           model_params,
+                                           model_state,
+                                           hyperparameters,
+                                           opt_init_rng)
+
+  logging.info('Initializing metrics bundle.')
   if tokenizer_vocab_path:
     workload.init_metrics_bundle(tokenizer_vocab_path)
-  wandb.init()
+  if wandb is not None:
```
runame, line 223: When using PyTorch DDP, should wandb be initialized in all python processes? If not, we can add a `RANK == 0` condition here.

runame (approved):
LGTM! Does the reference submission test already work for the PyTorch workloads?

It looks like yapf still has to be fixed.

submission_runner.py
znado: Aha yep, added that.

PR #135, opened 2022-09-18
conformer-pytorch

chandramouli-sastry:
Few notes:
- This implementation differs from the jax implementation in terms of the BatchNorm: 
       - The scale parameter of the pytorch batchnorm is initialized with 1s whereas the jax implementation uses (1s + learnable vector initializes 0s) as the scale parameter. 
       - The pytorch batchnorm is synchronized throughout training whereas jax synchronizes batchnorm only before the evaluation cycles. 
-   The query scaler in pytorch is applied to the vector before deriving the query vector, whereas jax applies the scaler after deriving the query vector. 
- The number of parameters between pytorch and jax are equal only for num_attention_heads=1. It seems that the way biases are set up in jax for attention-heads>1 is different than in pytorch. I am yet to examine this further. 
- Each of the pytorch modules are tested against the corresponding jax modules to ensure identical outputs for identical inputs and weights. I have run the code in both dataparallel to ddp modes and confirmed that they run correctly. 
- The full end-to-end training will help identify gaps, if any. 
- Some other notes:
    - The dictionary defined on line 38 in "prepare_data.py" does not contain 'test-other'. 
    - When running in ddp, I had to comment out wandb.init().


algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
```diff
@@ -0,0 +1,424 @@
...
+        key=outputs,
+        value=outputs,
+        key_padding_mask=paddings,
+        need_weights=False,
+    )
+    outputs = self.dropout(outputs)
+    return outputs
+
+
+class BatchNorm(nn.Module):
```
znado, line 281: in order to make the scale init the same as jax, can we instead make another class that subclasses `nn.BatchNorm1d`, where the only thing it does is override `reset_parameters` and `forward`? looking at the [source](https://pytorch.org/docs/stable/_modules/torch/nn/modules/batchnorm.html#BatchNorm1d) I think that's all we need for it? `reset_parameters` would just init `weight` to 0, and `forward` would pass `(1 + self.weight)` to `F.batch_norm`, something like below:

```
    def reset_parameters(self) -> None:
        self.reset_running_stats()
        if self.affine:
            init.zeros_(self.weight)
            init.zeros_(self.bias)
```

```
    def forward(self, input: Tensor) -> Tensor:
        self._check_input_dim(input)

        # exponential_average_factor is set to self.momentum
        # (when it is available) only so that it gets updated
        # in ONNX graph when this node is exported to ONNX.
        if self.momentum is None:
            exponential_average_factor = 0.0
        else:
            exponential_average_factor = self.momentum

        if self.training and self.track_running_stats:
            # TODO: if statement only here to tell the jit to skip emitting this when it is None
            if self.num_batches_tracked is not None:  # type: ignore[has-type]
                self.num_batches_tracked.add_(1)  # type: ignore[has-type]
                if self.momentum is None:  # use cumulative moving average
                    exponential_average_factor = 1.0 / float(self.num_batches_tracked)
                else:  # use exponential moving average
                    exponential_average_factor = self.momentum

        r"""
        Decide whether the mini-batch stats should be used for normalization rather than the buffers.
        Mini-batch stats are used in training mode, and in eval mode when buffers are None.
        """
        if self.training:
            bn_training = True
        else:
            bn_training = (self.running_mean is None) and (self.running_var is None)

        r"""
        Buffers are only updated if they are to be tracked and we are in training mode. Thus they only need to be
        passed when the update should occur (i.e. in training mode when they are tracked), or when buffer stats are
        used for normalization (i.e. in eval mode when buffers are not None).
        """
        return F.batch_norm(
            input,
            # If buffers are not to be tracked, ensure that they won't be updated
            self.running_mean
            if not self.training or self.track_running_stats
            else None,
            self.running_var if not self.training or self.track_running_stats else None,
            (1.0 + self.weight),
            self.bias,
            bn_training,
            exponential_average_factor,
            self.eps,
        )
```

I'm not torch expert tho, so maybe I'm missing something.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
```diff
@@ -0,0 +1,424 @@
...
+from dataclasses import dataclass
+import math
+from typing import Tuple
+
+import torch
+from torch import nn
+import torch.nn.functional as F
+
+from . import preprocessor
+from .spectrum_augmenter import SpecAug
```
znado, line 14: can we use absolute imports instead of relative? that way if we reorganize the dir structure in the future we don't break these

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
```diff
@@ -0,0 +1,424 @@
...
+    input_length = paddings.shape[1]
+    stride = self.filter_stride[0]
+    pad_len = (input_length + stride - 1) // stride * stride - input_length
+    out_padding = F.conv1d(
+        input=torch.cat([
+            paddings[:, None, :],
+            torch.zeros(
+                size=(paddings.shape[0], 1, pad_len), device=paddings.device)
+        ],
+                        dim=2),
```
znado, line 173: the way that this line breaks make it a bit hard to follow, can you instead pull out the input=... value to a python var on the line above, and pass it in like input=out_padding_input (or whatever you wanna call it)

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
```diff
@@ -0,0 +1,424 @@
...
+    self.max_timescale = max_timescale
+    self.embedding_dim = embedding_dim
+    num_timescales = self.embedding_dim // 2
+    log_timescale_increment = math.log(
+        float(self.max_timescale) / float(self.min_timescale)) / (
+            num_timescales - 1)
+    inv_timescales = self.min_timescale * \
+        torch.exp(torch.arange(num_timescales, dtype=torch.float32)
+                  * -log_timescale_increment)
+    self.register_buffer("inv_timescales", inv_timescales[None, None, :])
```
znado, line 225: nit: here and elsewhere, use single quotes just so we're consistent with the rest of the codebase

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
```diff
@@ -0,0 +1,424 @@
...
+class ConformerBlock(nn.Module):
+
+  def __init__(self, config: ConformerConfig):
+    super().__init__()
+
+    self.ff1 = FeedForwardModule(config)
+    self.mhsa = MultiHeadedSelfAttention(config)
+    self.conv = ConvolutionBlock(config)
+    self.ff2 = FeedForwardModule(config)
+
```
znado, line 357: nit: delete unnecessary newline

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/preprocessor.py
```diff
@@ -0,0 +1,557 @@
...
+    ])
+
+
+def _hertz_to_mel(frequencies_hertz):
+  """Convert hertz to mel."""
+  if type(frequencies_hertz) in [type(0.0), type(0)]:
+    return _MEL_HIGH_FREQUENCY_Q * math.log(1.0 + (frequencies_hertz /
+                                                   _MEL_BREAK_FREQUENCY_HERTZ))
+  return _MEL_HIGH_FREQUENCY_Q * torch.log(1.0 + (frequencies_hertz /
+                                                  _MEL_BREAK_FREQUENCY_HERTZ))
```
znado, line 210: nit: this could probably be a bit cleaner like:

```
log_fn = math.log if type(frequencies_hertz) in [type(0.0), type(0)] else torch.log
return _MEL_HIGH_FREQUENCY_Q * log_fn(1.0 + (frequencies_hertz /
                                                  _MEL_BREAK_FREQUENCY_HERTZ))
```

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/spectrum_augmenter.py
```diff
@@ -0,0 +1,160 @@
...
+        multiplicity=multiplicity,
+        max_ratio=1.0,
+        device=inputs.device)
+
+    outputs = torch.einsum('bxy,by->bxy', inputs, block_arrays)
+    return outputs
+
+  def forward(self, inputs, paddings):
+    lengths = torch.einsum('bh->b', 1 - paddings).long()
+
```
znado, line 156: nit: delete unnecessary newline (here and below)

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+from algorithmic_efficiency import param_utils
+from algorithmic_efficiency import spec
+from algorithmic_efficiency.pytorch_utils import pytorch_setup
+import algorithmic_efficiency.random_utils as prng
+from algorithmic_efficiency.workloads.librispeech_conformer import metrics
+from algorithmic_efficiency.workloads.librispeech_conformer import workload
+
+from .model import ConformerConfig
+from .model import ConformerEncoderDecoder
+from .model import initialize
```
znado, line 21: nit: change to absolute imports

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+
+class LibriSpeechConformerWorkload(workload.BaseLibrispeechWorkload):
+
+  def init_model_fn(self, rng: spec.RandomState) -> spec.ModelInitState:
+    torch.random.manual_seed(rng[0])
+    model = ConformerEncoderDecoder(ConformerConfig())
+    self._model = model
+    self.ctc_loss = torch.nn.CTCLoss(blank=0, reduction="none")
+    # Run model once to initialize lazy layers
+    t = 320000
```
znado, line 34: given that this is used below as well, I would pull this into a file level constant like `MAX_INPUT_LENGTH = 320000` and replace all uses of it

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+
+  def build_input_queue(self,
+                        data_rng: jax.random.PRNGKey,
+                        split: str,
+                        data_dir: str,
+                        global_batch_size: int,
+                        num_batches: Optional[int] = None,
+                        repeat_final_dataset: bool = False):
+    per_device_batch_size = int(global_batch_size / N_GPUS)
+    n_inputs = 4
```
znado, line 102: nit: delete this and below use `enumerate(keys)`

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+            device=DEVICE)
+        dist.broadcast(tensor, src=0)
+        # Note that the order of the keys is important.
+        tensors = tensor.split([320000, 320000, 256, 256], dim=-1)
+        batch = {}
+        for key, n in zip(keys, range(n_inputs)):
+          batch[key] = tensors[n][RANK]
+      yield batch
+
+  def compute_loss(self, logits, logit_paddings, targets, target_paddings):
```
znado, line 150: is this supposed to be `loss_fn` ?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+    logprobs = torch.log_softmax(logits, dim=-1)
+    input_lengths = torch.einsum('bh->b', 1 - logit_paddings).long()
+    target_lengths = torch.einsum('bh->b', 1 - target_paddings).long()
+    per_seq_loss = self.ctc_loss(
+        logprobs.permute(1, 0, 2),
+        targets.long(),
+        input_lengths,
+        target_lengths)
+    average_loss = per_seq_loss / target_lengths.sum()
```
znado, line 159: (we need to do a pass over the codebase to standardize this sort of logic, I know we do this in the jax workload, so this is an issue with both, but I'll talk to Sourabh about if I'm mistaken)

instead of returning the average here, do we instead need to make a helper function that returns the per_seq_loss and summed target len, and then inside `_eval_model_on_split` we would call the new helper function and sum up the numerator and denominator before dividing? (then `loss_fn` would just call the helper and immediately return the division)

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+          model_state,
+          spec.ForwardPassMode.EVAL,
+          model_rng,
+          update_batch_norm=False)
+      decoded, decoded_paddings = self.greedy_decode(logits, logits_padding)
+      word_errors, num_words = metrics.compute_wer(
+          decoded=decoded.detach().cpu().numpy(),
+          decoded_paddings=decoded_paddings.detach().cpu().numpy(),
+          targets=batch['targets'].detach().cpu().numpy(),
+          target_paddings=batch["target_paddings"].detach(
```
znado, line 230: single quotes

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+              num_words
+      }
+      total_metrics = {
+          k: v + batch_metrics[k] for k, v in total_metrics.items()
+      }
+    if USE_PYTORCH_DDP:
+      for metric in total_metrics.values():
+        dist.all_reduce(metric)
+    return {
+        "ctc_loss": float(total_metrics["ctc_loss"].item() / num_examples),
```
znado, line 251: if my above comment about numerator / denominator is mistaken, and we keep 'ctc_loss' as a key, I think this needs to be ` / num_batches` not num_examples?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -0,0 +1,253 @@
...
+        "wer": total_metrics["word_errors"] / total_metrics["num_words"]
```
znado, line 252: nit: add trailing comma

reference_submissions/librispeech_conformer/librispeech_pytorch/submission.py
```diff
@@ -74,8 +87,8 @@ def data_selection(workload: spec.Workload,
                    rng: spec.RandomState) -> Dict[str, spec.Tensor]:
   """Select data from the infinitely repeating, pre-shuffled input queue.
 
-  Each element of the queue is a batch of training examples and labels.
-  """
+    Each element of the queue is a batch of training examples and labels.
+    """
```
znado, line 91: nit: delete indent

tests/test_num_params.py
```diff
@@ -30,7 +39,15 @@
...
 
-WORKLOADS = ['mnist', 'cifar', 'imagenet_resnet', 'imagenet_vit', 'wmt', 'ogbg']
+WORKLOADS = [
+    'mnist',
+    'cifar',
+    'imagenet_resnet',
+    'imagenet_vit',
+    'wmt',
+    'ogbg',
+    'librispeech_conformer'
```
znado, line 49: nit: add trailing comma

znado (commented):
LGTM once comments addressed. Sourabh should also review some of the more complicated speech logic for correctness.

Thank you so much for this, I'm very excited to get this in!

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: (I also see that I have set `reduction='none'` for the ctc-loss and haven't summed the `per_seq_loss` before computing the average_loss -- will fix this!)

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: In order to resolve style errors related to changing argument names in subclass implementation, I renamed `loss_fn` to `compute_loss` for this workload and totally forgot to add a dummy method to override the abstract method `loss_fn` of the super class: what do you suggest we do?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: Good catch! But, I think the denominator has to be the number of gpus because each gpu has the average loss for the batch slice it received? I'll clean all this up soon!

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
chandramouli-sastry: Nice, will do this!

tests/test_num_params.py
chandramouli-sastry: Done.

reference_submissions/librispeech_conformer/librispeech_pytorch/submission.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: I cleaned this up by returning a dictionary from the compute_loss function: i have only implemented a dummy `loss_fn` as of now.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/preprocessor.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/spectrum_augmenter.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/model.py
chandramouli-sastry: Done.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
znado: you can leave it as is for now and we can refactor it later

(I had refactored the data iterator for librispeech to produce tuple(input, input_padding) and tuple(target, target_padding), and then loss_fn can take in logits_batch=tuple(logit, logit_padding) and targets_batch=(targets, target_padding), but we can fix that once this is checked in)

znado (approved):
LGTM pending Sourabh's librispeech logic review

znado:
We should probably remedy the two diffs you highlighted in the PR before you start an end-to-end run:
- The query scaler in pytorch is applied to the vector before deriving the query vector, whereas jax applies the scaler after deriving the query vector.
- The number of parameters between pytorch and jax are equal only for num_attention_heads=1. It seems that the way biases are set up in jax for attention-heads>1 is different than in pytorch. I am yet to examine this further.

Not sure about the biases, but we have seen massive model stability/perf changes depending on how the query scaling is applied. Can we implement a custom (subclassed? forked?) attention layer in python for pytorch that behaves the same as flax, or is this behavior baked into the c++ code? Or is it possible to modify what we pass in/get back from the c++ code, like we did with BN?

chandramouli-sastry:
> We should probably remedy the two diffs you highlighted in the PR before you start an end-to-end run:
> 
> * The query scaler in pytorch is applied to the vector before deriving the query vector, whereas jax applies the scaler after deriving the query vector.
> * The number of parameters between pytorch and jax are equal only for num_attention_heads=1. It seems that the way biases are set up in jax for attention-heads>1 is different than in pytorch. I am yet to examine this further.
> 
> Not sure about the biases, but we have seen massive model stability/perf changes depending on how the query scaling is applied. Can we implement a custom (subclassed? forked?) attention layer in python for pytorch that behaves the same as flax, or is this behavior baked into the c++ code? Or is it possible to modify what we pass in/get back from the c++ code, like we did with BN?

I fixed the query-scaling issue by scaling the bias as well. 

This also fixes the parameter count mismatch issue: the mismatch was in-fact coming from the difference in the query-scaler implementations but I wrongly thought that the issue was with the way the biases were set up (because these parameters were also 1D vectors!). Specifically, the jax code shares the same scaler across all the heads whereas I had set up the pytorch code such that each head got its own scaler; this is now rectified. 

Also: Scaling the bias and query vectors such that the query-scaler is shared across all heads requires us to know how the query is split across the heads: (nheads, embed_dim//nheads) or (embed_dim//nheads, nheads). I identified that it was the former with a few simple tests and followed this while applying query-scaling.

PR #139, opened 2022-09-23
Update setup.cfg

priyakasimbeg:
Add wandb to installation under `install_requires`.


fsschneider:
Do we require `wandb`?

znado:
I think we talked and agreed to add it, but on second thought, lets keep it optional for now, and let users install it if they want. We already have the try/except around importing it if it isn't there.

PR #140, opened 2022-09-24
Fix reference submission tests

runame:
Fixes the ref sub tests with PyTorch DDP, verified that it works for WMT and OGBG.


znado (approved):
thanks so much for fixing this!

Issue #137, opened 2022-09-21
fastmri workload not initally running on jax 


rakshithvasudev:
It seems like the tf input pipeline seems to have issues? I couldn't debug much. I ensured that there is dataset in the `/dataset` folder and there are no file permission issues and passed `data_dir `argument as seen below.  

example: 
```
mlcommons@3d29526317fc:~/algorithmic-efficiency$ ls -ltr /dataset/knee_singlecoil_train/file1000492.h5
-rw-rw-r-- 1 1002 1002 86389120 Nov 23  2018 /dataset/knee_singlecoil_train/file1000492.h5
```


```
2022-09-21 21:49:59.097301: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F AVX512_VNNI F
MA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
Traceback (most recent call last):
  File "submission_runner.py", line 450, in <module>
    app.run(main)
  File "/home/mlcommons/.local/lib/python3.8/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)                                                                                                                                                                                                                                                    File "/home/mlcommons/.local/lib/python3.8/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))                                                                                                                                                                                                                                                     File "submission_runner.py", line 429, in main
    score = score_submission_on_workload(workload,
  File "submission_runner.py", line 361, in score_submission_on_workload
    timing, metrics = train_once(workload, global_batch_size,                                                                                                                                                                                                                File "submission_runner.py", line 243, in train_once
    batch = data_selection(workload,
  File "/home/mlcommons/algorithmic-efficiency/reference_submissions/fastmri/fastmri_jax/submission.py", line 142, in data_selection
    return next(input_queue)
  File "/home/mlcommons/algorithmic-efficiency/algorithmic_efficiency/workloads/fastmri/fastmri_jax/input_pipeline.py", line 224, in load_fastmri_split
    ds = ds.enumerate().map(process_example, num_parallel_calls=16)
  File "/home/mlcommons/.local/lib/python3.8/site-packages/tensorflow/python/data/ops/dataset_ops.py", line 2050, in map
    return ParallelMapDataset(
[...]

PR #141, opened 2022-09-24
Fastmri jax fixes

priyakasimbeg:
Fixes for fastmri JAX dataset:

- fix `weights` definition; define weights after batch_size is known
- fix `load_fastmri_split` method so that it returns iterators instead of potentially acting as an iterator
- fix `batch_size` from 512 (which causes OOM) to 64
- fix error w. attempted numpy conversion in data_utils function; check if data is already of numpy.ndarray type
- fix `_pad_zeros_like` method to return the padded batch instead of the padding

Details in https://github.com/mlcommons/algorithmic-efficiency/issues/137


znado (commented):
It really runs out of memory after a bs of 8? I ran with up to 256 (using fake random data) with up to 256, using `tests/reference_submission_test.py`.

znado (approved):
LGTM!

PR #142, opened 2022-09-26
Imagenet v2

znado:
Adding 'imagenet_v2/matched-frequency:3.0.0' as a test set for ImageNet. Uses tf.data for both jax and pytorch, and caches all images in memory.

Question: is it ok to call `itertools.cycle` (which saves a copy of an iterator in python to be run through endlessly) after moving the test images to the GPU, or will that always reserve them in GPU memory? Is that OK, or should we move them to GPU each eval?

### DIFF
--- a/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
@@ -41,12 +41,10 @@ class ImagenetResNetWorkload(BaseImagenetResNetWorkload):
                      data_dir: str,
                      batch_size: int,
                      cache,
-                     repeat_final_dataset,
-                     num_batches):
+                     repeat_final_dataset):
     del data_rng
     del cache
     del repeat_final_dataset
-    del num_batches
     is_train = split == 'train'
 
     eval_transform_config = transforms.Compose([
--- a/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
@@ -35,19 +35,18 @@ class ImagenetResNetWorkload(BaseImagenetResNetWorkload):
       self._param_types = param_utils.pytorch_param_types(self._param_shapes)
     return self._param_types
 
-  def build_input_queue(self,
-                        data_rng: spec.RandomState,
-                        split: str,
-                        data_dir: str,
-                        global_batch_size: int):
-    return self._build_dataset(data_rng, split, data_dir, global_batch_size)
-
   def _build_dataset(self,
                      data_rng: spec.RandomState,
                      split: str,
                      data_dir: str,
-                     batch_size: int):
+                     batch_size: int,
+                     cache,
+                     repeat_final_dataset,
+                     num_batches):
     del data_rng
+    del cache
+    del repeat_final_dataset
+    del num_batches
     is_train = split == 'train'
 
     eval_transform_config = transforms.Compose([
--- a/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
@@ -1,10 +1,12 @@
 """ImageNet workload implemented in PyTorch."""
 
 import contextlib
+import itertools
 import math
 import os
 from typing import Dict, Tuple
 
+import numpy as np
 import torch
 from torch import nn
 import torch.distributed as dist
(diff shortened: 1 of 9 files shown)


algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
```diff
@@ -88,10 +100,10 @@ def _build_dataset(self,
       else:
         sampler = data_utils.DistributedEvalSampler(
             dataset, num_replicas=N_GPUS, rank=RANK, shuffle=False)
-      batch_size //= N_GPUS
+      per_device_batch_size = global_batch_size // N_GPUS
     dataloader = torch.utils.data.DataLoader(
         dataset,
-        batch_size=batch_size,
+        batch_size=per_device_batch_size,
```
runame, line 106: `per_device_batch_size` is not defined if `USE_PYTORCH_DDP` is False.

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_v2.py
```diff
@@ -0,0 +1,62 @@
...
+
+import jax
+import numpy as np
+import tensorflow_datasets as tfds
+
+from algorithmic_efficiency.workloads.imagenet_resnet.imagenet_jax import \
+    input_pipeline
+
+
+def _shard(x):
```
runame, line 15: `algorithmic_efficiency.data_utils.shard_numpy_ds` could be adjusted to be also used here.

submission_runner.py
```diff
@@ -108,7 +108,10 @@
 flags.DEFINE_integer('num_tuning_trials',
                      20,
                      'The number of external hyperparameter trials to run.')
-flags.DEFINE_string('data_dir', '~/tensorflow_datasets/', 'Dataset location')
+flags.DEFINE_string('data_dir', '~/tensorflow_datasets/', 'Dataset location.')
+flags.DEFINE_string('imagenet_v2_data_dir',
```
runame, line 112: Why do we need to specify a separate directory for this dataset? Can't we place it with a fixed naming convention within `data_dir`?

runame (approved):
Good question regarding `itertools.cycle()`... I'm not sure, but I would guess that they keep the data in GPU memory, since I think they just create a new list which stores the elements returned by the iterator. This might create problems for memory-hungry algorithms during training, so it might be safer to just transfer the arrays to GPU each evaluation. Did you test if something like Adam runs OOM?

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_v2.py
znado: I was going to use that but we convert these to numpy first so the `x = x._numpy()` code will fail in there. But we absolutely need to clean up how we do tf.tensor/np.ndarray/torch.tensor/jax.devicearray handling/padding/conversion with datasets, I plan on doing that once the fires have been put out with implementing them all

submission_runner.py
znado: yeah, I wasn't sure how we wanted to do this, because for imagenet this is the only workload where we really have two datasets (one for train/val and one for test) which come from different sources. the rest use 1 dataset, or with WMT they use two (WMT 14 vs 17) but they are both in the same format via tfds. if we were only using the `tf.data` pipeline _or_ the pytorch one I think we could do this, but I think it will be easier and more explicit to let the user specify it (even if it is more annoying)

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_v2.py
runame: Yeah it would need to be adjusted first. And yes, we'll have to clean up related stuff anyway ;)

submission_runner.py
runame: My reasoning was that we probably want to automate the data downloading process anyway, so for most users we can determine the location of the imagenet test dataset within the user specified data dir anyway. But I'm ok with leaving it as it is for now.

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
znado: ah good catch, thanks!

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
```diff
@@ -26,6 +29,18 @@
 USE_PYTORCH_DDP, RANK, DEVICE, N_GPUS = pytorch_setup()
 
 
+def imagenet_v2_to_torch(batch):
+  # Slice off the part of the batch for this device and then transpose from
+  # [N, H, W, C] to [N, C, H, W]. Only transfer the inputs to GPU.
+  new_batch = {}
+  for k, v in batch.items():
+    new_v = v[RANK]
```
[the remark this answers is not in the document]
znado: Got it, I passed a flag down so that we never shard in the first place if `USE_PYTORCH_DDP=True`, which I think should be equivalent.

znado:
Swapped the ordering of `imagenet_v2_to_torch` and `itertools.cycle` so that itertools stores the numpy-only iterator.

PR #143, opened 2022-09-27
Add mixup option to ImageNet workloads

runame:
Adds mixup data augmentation option to the ImageNet workloads. It can be selected by specifying the hyperparameters `use_mixup` and optionally `mixup_alpha`. As discussed I went with the less efficient but simpler version where we take the convex combination over the one-hot labels instead over two losses.

### DIFF
--- a/algorithmic_efficiency/data_utils.py
+++ b/algorithmic_efficiency/data_utils.py
@@ -29,7 +29,7 @@ def shard_numpy_ds(xs):
   return jax.tree_map(_prepare, xs)
 
 
-def mixup(batch, alpha=0.1):
+def mixup_pytorch(batch, alpha=0.1):
   inputs, targets = batch
   # Transform to one-hot targets.
   targets = F.one_hot(targets, num_classes=1000)
@@ -55,7 +55,7 @@ def cycle(iterable,
       batch = next(iterator)
       if use_mixup:
         assert keys == ('inputs', 'targets')
-        batch = mixup(batch, alpha=mixup_alpha)
+        batch = mixup_pytorch(batch, alpha=mixup_alpha)
       assert len(keys) == len(batch)
       yield dict(zip(keys, batch))
     except StopIteration:
--- a/algorithmic_efficiency/data_utils.py
+++ b/algorithmic_efficiency/data_utils.py
@@ -29,7 +29,7 @@ def shard_numpy_ds(xs):
   return jax.tree_map(_prepare, xs)
 
 
-def mixup_pytorch(batch, alpha=0.1):
+def mixup_pytorch(batch, alpha=0.2):
   inputs, targets = batch
   # Transform to one-hot targets.
   targets = F.one_hot(targets, num_classes=1000)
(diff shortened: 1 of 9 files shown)


algorithmic_efficiency/data_utils.py
```diff
@@ -28,13 +29,33 @@ def _prepare(x):
   return jax.tree_map(_prepare, xs)
 
 
+def mixup(batch, alpha=0.1):
```
znado, line 32: can you call this `mixup_pytorch`?

submission_runner.py
```diff
@@ -196,12 +196,20 @@ def train_once(
 
   # Workload setup.
   logging.info('Initializing dataset.')
+  use_mixup = (
+      hyperparameters.use_mixup
+      if hasattr(hyperparameters, 'use_mixup') else False)
+  mixup_alpha = (
+      hyperparameters.mixup_alpha
+      if hasattr(hyperparameters, 'mixup_alpha') else 0.1)
```
znado, line 204: we will never tune these, because the competition rules prohibit data aug changing on a per-workload basis, so I think we can remove all the changes to submision_runner.py

algorithmic_efficiency/spec.py
```diff
@@ -111,7 +111,9 @@ def build_input_queue(self,
                         global_batch_size: int,
                         cache: Optional[bool] = None,
                         repeat_final_dataset: Optional[bool] = None,
-                        num_batches: Optional[int] = None) -> Dict[str, Any]:
+                        num_batches: Optional[int] = None,
+                        use_mixup: bool = False,
+                        mixup_alpha: float = 0.1) -> Dict[str, Any]:
```
znado, line 116: (I mention it also below, but) I think we should remove the changes to build_input_queue, because we will always run with a fixed mixup_alpha=0.2 only for imagenet_vit, and submitters aren't allowed to tune it. then we can just pass use_mixup=true only for the training split (see below comments)

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/workload.py
```diff
@@ -30,7 +30,9 @@ def _build_dataset(self,
                      data_dir: str,
                      global_batch_size: int,
                      cache: Optional[bool] = None,
-                     repeat_final_dataset: Optional[bool] = None):
+                     repeat_final_dataset: Optional[bool] = None,
+                     use_mixup: bool = False,
+                     mixup_alpha: float = 0.1):
```
znado, line 35: we can remove mixup_alpha from here, and then in the imagenet_vit base class, we have a 2 line `_build_dataset` function that just passes `use_mixup=(split == 'train')` to the super class' implementation (same for pytorch)

znado (approved):
LGTM, I just think we can remove some of the changes to make it even less generic/tunable. Feel free to merge once the comments are resolved.

runame:
Ok makes sense, I have followed your suggestions and restricted mixup to be only used with ImageNet-ViT.

PR #147, opened 2022-09-28
WandB Tracking Suggestions 

pomonam:
I have added a few more lines of code that I think is a good practice when using wandb tracking. 

 - I made wandb to be an optional feature and can be enabled with a flag `--wandb`. 
 - For the default behaviour when using wandb, we save the configuration (`flags.FLAGS`) and log intermediate results (should work for all workloads). 
 - Exit with `wandb.finish()`.

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -117,7 +117,8 @@ flags.DEFINE_enum(
     help='Whether to use Jax or Pytorch for the submission. Controls among '
     'other things if the Jax or Numpy RNG library is used for RNG.')
 flags.DEFINE_boolean('profile', False, 'Whether to produce profiling output.')
-flags.DEFINE_boolean('wandb', False,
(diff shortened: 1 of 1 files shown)


submission_runner.py
```diff
@@ -426,6 +436,10 @@ def main(_):
                                        FLAGS.tokenizer_vocab_path)
   logging.info('Final %s score: %f', FLAGS.workload, score)
 
+  if FLAGS.wandb and wandb is not None and RANK == 0:
+    wandb.log({"score": score})
```
runame, line 440: Nit: single quotation marks for consistency.

runame (approved):
Nice, thanks!

submission_runner.py
pomonam: Thank you! Fixed :)

znado (approved):

PR #151, opened 2022-10-06
Unify logging for all workloads

pomonam:
Before making a PR that checkpoints, I am first making the logging PR available. Many functions are borrowed from #60.

It creates a directory given by `--experiment_dir` and stores all results inside. It includes meta-data, flags, hyperparameters, and intermediate results. Additionally, if available, it stores TF files and users can view the results in wandb. I also made it easy to extend the code if the user wants to use their own loggers. 

One minor question: the logging requires `clu` package in `setup.cfg`. Should we make this package required in general?

In the upcoming PRs, the followings need to be done:
- Implement checkpointing
- Update README.md once we are satisfied with the logging logic.


submission_runner.py
```diff
@@ -8,7 +8,8 @@
     --submission_path=reference_submissions/mnist/mnist_jax/submission.py \
     --tuning_ruleset=external \
     --tuning_search_space=reference_submissions/mnist/tuning_search_space.json \
-    --num_tuning_trials=3
+    --num_tuning_trials=3 \
+    --experiment_dir=experiments
```
znado, line 12: I would make the example have an absolute dir, even if people need to change it, so something like `/home/pomonam/mlcommons`

that way there isn't any confusion about where the dir is relative to

submission_runner.py
```diff
@@ -116,17 +113,16 @@
...
-flags.DEFINE_string('summary_log_dir',
-                    '',
-                    'Location to dump tensorboard summaries.')
 flags.DEFINE_string('tokenizer_vocab_path',
                     '',
                     'Location to read tokenizer from.')
+
+flags.DEFINE_string('experiment_dir',
+                    None,
+                    'The root directory to store all experiments')
```
znado, line 122: I would specify here that while not required, the dir should be an absolute dir because otherwise it will be relative to wherever the user is running the script from

submission_runner.py
```diff
@@ -197,6 +193,27 @@ def train_once(
 ) -> Tuple[spec.Timing, spec.Steps]:
   data_rng, opt_init_rng, model_init_rng, rng = prng.split(rng, 4)
 
+  # Logger setup.
+  logging.info('Initializing logger.')
+  metrics_logger = None
+  if log_dir is not None:
+    hparams_fname = os.path.join(log_dir, 'hparams.json')
+    meta_data = get_meta_data(workload)
+    meta_fname = os.path.join(log_dir, 'meta_data.json')
```
znado, line 202: nit: I prefer `filename` instead of `fname` because the latter can imply function or something else :)

submission_runner.py
```diff
@@ -441,4 +472,5 @@ def main(_):
 
 
 if __name__ == '__main__':
+  flags.mark_flag_as_required('experiment_dir')
```
znado, line 475: we should also mark the following as required:
workload
framework
submission_path
tuning_ruleset
tuning_search_space

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
```diff
@@ -287,13 +287,13 @@ def _eval_model_on_split(self,
...
-      self.summary_writer.scalar('{}/ctc_loss'.format(split),
-                                 computed_metrics['ctc_loss'],
-                                 global_step)
+    # if self.summary_writer is not None:
+    #   self.summary_writer.scalar('{}/WER'.format(split),
+    #                              computed_metrics['wer'],
+    #                              global_step)
+    #   self.summary_writer.scalar('{}/ctc_loss'.format(split),
+    #                              computed_metrics['ctc_loss'],
+    #                              global_step)
```
znado, line 296: delete

znado (approved):
Looks great, thanks for cleaning this up!

I think it's fine to make `clu` required in general.

pomonam:
Thank you a lot for the feedback! The most recent commit should address all concerns.

znado (approved):

PR #152, opened 2022-10-07
DLRM Remove AUC PyTorch

rakshithvasudev:


znado:
Thanks so much for cleaning this up, these sorts of PRs are going to be increasingly useful as we get closer to the final version of the code!

znado (approved):

PR #156, opened 2022-10-08
FastMRI and WMT test sets

znado:
As discussed with Mike, because the official "test" set does not have labels, we will use the first 100 h5 files of the validation set for validation and the remaining 99 files as the test set.

Adding the correct number of examples for the WMT test set.

### DIFF
--- a/algorithmic_efficiency/workloads/fastmri/workload.py
+++ b/algorithmic_efficiency/workloads/fastmri/workload.py
@@ -34,11 +34,11 @@ class BaseFastMRIWorkload(spec.Workload):
 
   @property
   def num_validation_examples(self):
-    return 7135
+    return 3554
 
   @property
   def num_test_examples(self):
-    return None
+    return 3581
 
   @property
   def train_mean(self):
--- a/algorithmic_efficiency/workloads/wmt/workload.py
+++ b/algorithmic_efficiency/workloads/wmt/workload.py
@@ -51,12 +51,13 @@ class BaseWmtWorkload(spec.Workload):
 
   @property
   def num_validation_examples(self):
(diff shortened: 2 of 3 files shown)


algorithmic_efficiency/workloads/fastmri/workload.py
```diff
@@ -34,11 +34,11 @@ def num_eval_train_examples(self):
 
   @property
   def num_validation_examples(self):
-    return 7135
+    return 3554
```
runame, line 37: I think we haven't discussed yet how we should handle the case where `(num_{eval_train, validation, test}_examples % global_batch_size) % N_GPU != 0`.

algorithmic_efficiency/workloads/wmt/workload.py
```diff
@@ -51,12 +51,13 @@ def num_eval_train_examples(self):
...
   def num_validation_examples(self):
-    # wmt14_translate/de-en 'validation' split size
+    # wmt14_translate/de-en 'validation' split size.
     return 3000
 
   @property
   def num_test_examples(self):
-    return None
+    # wmt14_translate/de-en 'test' split size.
+    return 3003
```
runame, line 60: See comment above.

runame (approved):
LGTM, I guess we can decide on a solution to the issue mentioned in the comments after merging this PR.

algorithmic_efficiency/workloads/fastmri/workload.py
znado: yeah, we will need to support a padded final batch, which I think we sort of do right now, but we need to make sure if consistent. the batches all already have a 'weights' key that we can use to mask out any padding. I'll open an issue to make sure we don't forget.

Issue #157, opened 2022-10-08
Clean up param_shapes and param_types


znado:
Currently we have a lot of duplicate functions that set/get `param_shapes` and `param_types`. For example, these are practically identical across all jax/pytorch workloads:
jax:
```python
  @property
  def model_params_types(self):
    if self._param_shapes is None:
      raise ValueError(
          'This should not happen, workload.init_model_fn() should be called '
          'before workload.param_shapes!')
    if self._param_types is None:
      self._param_types = param_utils.jax_param_types(
          self._param_shapes.unfreeze())
    return self._param_types
```
pytorch:
```python
  @property
  def model_params_types(self):
    """The shapes of the parameters in the workload model."""
    if self._param_types is None:
      self._param_types = param_utils.pytorch_param_types(self._param_shapes)
    return self._param_types
```

and this is constant across both frameworks:
```python
  @property
  def param_shapes(self):
    if self._param_shapes is None:
      raise ValueError(
          'This should not happen, workload.init_model_fn() should be called '
          'before workload.param_shapes!')
    return self._param_shapes
```

 We should move these duplicate functions to the base `Workload` class, and add a basic `__init__` in the base class that sets
```python
    self._param_shapes = None
    self._param_types = None
```

We should also clean up the construction of `self._param_shapes`, as the same code is called in each `init_model_fn()`. We can add a new private method to the base class, `_make_param_shapes()`, which takes the model and a string indicating if it's jax or pytorch:
```python
def _make_param_shapes(self, framework, model):
  if framework == 'jax':
    return jax.tree_map(lambda x: spec.ShapeTuple(x.shape),
                                      params)
  elif framework == 'pytorch':
    return {
        k: spec.ShapeTuple(v.shape) for k, v in model.named_parameters()
    }
  else:
    ValueError
```

(it's probably preferable to return rather than set `self._param_shapes` to avoid side effects). Then we can call this from all `init_model_fn`'s, to make sure we have the same exact code.

Of course, we should also highlight any cases (if any) where this trend won't work.

Issue #160, opened 2022-10-08
Unit test for param_types


znado:
Similar to `tests/test_param_shapes.py`, we should have a `tests/test_param_types.py` which tests that the pytrees of types are the same for each workload in jax/pytorch.

PR #166, opened 2022-10-08
Refactor functions related to param_shapes and param_types

pomonam:
This PR address #157 

Specifically, the PR does the followings:
-  Add functions `pytorch_param_shapes` and `jax_param_shapes` in `param_utils.py` to unify the shape computation process.
- Creates `_param_shapes`, `_param_types`, and `_eval_iters` in the base workload and removes duplicate initializations.
- Unify properties `param_shapes` and `model_param_shapes` across all workloads. 
- Fixes code so that we assign the correct type in `param_types`.
- Add unit test for matching `model_param_types`.

### DIFF
--- a/algorithmic_efficiency/param_utils.py
+++ b/algorithmic_efficiency/param_utils.py
@@ -37,9 +37,9 @@ def jax_param_types(param_shapes, parent_name=''):
     else:
       if 'bias' in name:
         param_types_dict[name] = spec.ParameterType.BIAS
-      elif 'BatchNorm' in parent_name:
+      elif 'batchnorm' in parent_name.lower():
         param_types_dict[name] = spec.ParameterType.BATCH_NORM
-      elif 'Conv' in parent_name:
+      elif 'conv' in parent_name.lower():
         param_types_dict[name] = spec.ParameterType.CONV_WEIGHT
       # Note that this is exact equality, not contained in, because
       # flax.linen.Embed names the embedding parameter "embedding"
--- a/algorithmic_efficiency/param_utils.py
+++ b/algorithmic_efficiency/param_utils.py
@@ -1,4 +1,5 @@
 """Utilities for dealing with parameter-related logic like types and shapes."""
+import flax
 import jax
 
 from algorithmic_efficiency import spec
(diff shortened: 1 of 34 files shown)


pomonam:
I will also incorporate Issue #160 in this PR.

algorithmic_efficiency/param_utils.py
```diff
@@ -1,38 +1,45 @@
...
+    if isinstance(value, dict) or isinstance(value, flax.core.FrozenDict):
+      param_types_dict[name] = jax_param_types(value, parent_name=name)
     else:
       if 'bias' in name:
         param_types_dict[name] = spec.ParameterType.BIAS
-      elif 'BatchNorm' in name:
+      elif 'BatchNorm' in parent_name:
         param_types_dict[name] = spec.ParameterType.BATCH_NORM
-      elif 'Conv' in name:
+      elif 'Conv' in parent_name:
```
znado, line 42: not sure if this matters, but maybe to be safer we should do case insensitive comparisons? like:
```python
      elif 'batchnorm' in parent_name.lower():
        param_types_dict[name] = spec.ParameterType.BATCH_NORM
      elif 'conv' in parent_name.lower():
```

algorithmic_efficiency/workloads/mnist/mnist_pytorch/workload.py
```diff
@@ -124,6 +117,7 @@ def init_model_fn(self, rng: spec.RandomState) -> spec.ModelInitState:
     self._param_shapes = {
         k: spec.ShapeTuple(v.shape) for k, v in model.named_parameters()
     }
```
znado, line 119: it doesn't really matter because it's just the MNIST one, but should we replace this with
```python
self._param_shapes = param_utils.pytorch_param_shapes(model)
```

tests/test_param_types.py
```diff
@@ -0,0 +1,106 @@
...
+from algorithmic_efficiency.workloads.wmt.wmt_pytorch.workload import \
+    WmtWorkload as PyTorchWmtWorkload
+
+WORKLOADS = ['mnist', 'cifar', 'imagenet_resnet', 'imagenet_vit', 'wmt', 'ogbg']
+
+
+# Ideally we would match the types layer-wise, but for that we
+# have to ensure the exact same number of the types and that the
+# types of the weights of the same layer type actually match between
+# Jax and PyTorch, which is not always the case.
```
znado, line 39: is this because some layers fuse bias and weight together? can you put an example of the failure mode?

algorithmic_efficiency/param_utils.py
pomonam: That's a good point. Fixed!

pomonam:
Thank you! I modified the PR accordingly.

PR #169, opened 2022-10-10
Add general `output_activation_fn()` implementation

### LINKED ISSUE #158: Clean up `output_activation_fn`
Right now we have `output_activation_fn` unimplemented in basically every workload. Despite an incorrect code comment in `spec.py` which should be rewritten, this is supposed to take the model logits and turn them into probabilities. For FastMRI this should just be the identity. We can likely delete it from all non-base classes, and in the base class just implement an if/elif/else chain on the type of loss, because the logic should be the same across all workloads that share the same type of loss.

runame:
This PR resolves #158.

**Questions:**
- Currently, the function takes a `loss_type` argument. Why can't it just use `self.loss_type` instead, i.e. is there any use case for applying a different choice of output activation function?
- In `update_params` it also says that the `loss_type` keyword will determine the output activation function. Also, even if `self.loss_type` is used in `output_activation_fn`, it won't be used anywhere in the loss calculation, but only when the user explicitly calls the function. Moreover, it _can't_ be used in conjunction with `loss_fn`, since it requires logits.
- Not sure if there is a better way to select the framework besides using `FLAGS.framework`. Alternatively, we could add an additional argument, write Jax and PyTorch specific workload base classes, or set a `framework` property of the workload in the submission runner.

### DIFF
--- a/algorithmic_efficiency/spec.py
+++ b/algorithmic_efficiency/spec.py
@@ -5,12 +5,9 @@ import enum
 import functools
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
-from absl import flags
 from absl import logging
 import jax
 import torch.nn.functional as F
-
-FLAGS = flags.FLAGS
 
 
 class LossType(enum.Enum):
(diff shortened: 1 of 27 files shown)


algorithmic_efficiency/spec.py
```diff
@@ -229,7 +244,7 @@ def loss_fn(
       logits_batch: Union[Tuple[Tensor, Tensor], Tensor],
       mask_batch: Optional[Tensor] = None,
       label_smoothing: float = 0.0) -> Tensor:  # differentiable
-    """return oned_array_of_losses_per_example"""
+    """Return one_array_of_losses_per_example."""
```
znado, line 247: I think this should be "1-d array of losses per-example" ?

znado (approved):
Looks great!

- I think it's much cleaner to use `self.loss_type`, good point.
- Lets delete the "# This will define the output activation via `output_activation_fn`." comment. I think really loss_type here will just allow the submitter to branch on the type of loss function, they don't actually get any control over changing loss_type. I agree, the docs have been confusing for this! I don't know how much we've touched them in like 1+ years lol.
- I think it is bad practice/dangerous to use flags inside of library code, so I would almost rather modify the API to be `output_activation_fn(logits, framework)`. wdyt?

Also in `RULES.md` "The loss function will ... allowed to branch on the `loss_fn` enum/name." Let's change loss_fn there to loss_type.

runame:
Thanks for the feedback! Completely agree about using flags inside of the library code being bad practice. I went with the simplest fix and added an additional argument.

PR #171, opened 2022-10-10
Adding versions to tfds datasets.

znado:
Ensures that if TFDS updates their versions, we won't have the data redownloaded.


runame (approved):
LGTM

Issue #165, opened 2022-10-08
is_output_params defined as pass or NotImplementedError


pomonam:
In all workloads, `is_output_params` is set to either `pass` or `NotImplementedError`, i.e.
```
  # Return whether or not a key in spec.ParameterTree is the output layer
  # parameters.
  def is_output_params(self, param_key: spec.ParameterKey) -> bool:
    raise NotImplementedError
```
What is the correct implementation? I am assuming we need to provide a model-framework-specific key?

znado:
Yep we need to implement the definition in each workload, we should return if the key matches whatever the model has as the final layer

pomonam:
Sounds good. I have already implemented this. Once we get all PRs merged, I will make another PR that implements this :)

PR #172, opened 2022-10-11
Implement is_output_params

pomonam:
This PR addresses #165 for all workloads except for librispeech (since I thought it is likely to change). I also added fastmri and criteo1tb to the shape & type unit tests.


znado (approved):
Awesome, thanks so much for this, it's fantastic to get all our API implemented :)

Issue #164, opened 2022-10-08
Rename reference_submissions to reference_algos or something that isn’t “submission”


znado:
The current name is a misnomer, because the algos in `reference_submissions/` are not valid submissions because they have workload-dependent behavior.

PR #174, opened 2022-10-11
moving reference_submissions/ to tests/reference_algorithms/

znado:
Please let me know what you think of this renaming, I am very open to suggestions on other places/names for it.

1. The optimizers currently in `reference_submissions/` are not valid submissions, because they have workload-specific logic, so I'm renaming it to `reference_algorithms` instead.
2. I think we are soon going to outgrow the `reference_submissions/` dir as we make real baseline submissions, so we should de-emphasize it, which is why I moved it to `tests/`.

Fixes https://github.com/mlcommons/algorithmic-efficiency/issues/164


runame:
1. I really like the name change to `reference_algorithms`.
2. I slightly lean in favor of keeping it outside of `tests`, such that everything inside of `tests` is really just a unit test. But I'm also completely with moving it there if you think it's better this way.

pomonam:
I also like the name change `reference_algorithms`.

fsschneider:
I would have a slight preference for `reference_runs/` since one of its properties is that they do not have a proper search space but are rather specific runs. But `reference_algorithms` is also fine for me. In the medium term, we probably want to replace (and rename) them with the actual target-setting runs. Having those runs public in our repo would be helpful. We could, therefore, also directly call them `target_setting_runs/` (although they are just temporary target-setting runs.

I also have a slight preference for keeping it outside of the `tests/` directory for the reasons @runame mentioned.
Again, I would expect us to change this directory to `baselines/` in the near future. Having those baselines publically available would help submitters get started.

znado:
Ok, I moved the dir out of tests/ and just renamed it to reference_algorithms. I think we should stick with the convention of only using the word "submission" when we have a submission that is valid under the competition rules, but I am open to "algorithm" or "run" for anything else.

Currently we also have "target_setting_runs" so we should reconcile "run" vs "algorithm" for these two, maybe by merging "target_setting_runs" and "reference_algorithms" somehow. I think it's important to keep all the current submission.py files, but we could:
- make a new top level dir called "references" or something and have "reference_algorithms" and "target_setting_runs" below that
- keep the "reference_algorithms" top level dir name and make a new subdir called "target_setting_runs" below it (possibly moving its current contents to a new subdir too)

What do people think?

znado:
Ok, I moved `target_setting_runs => reference_algorithms/target_setting_algorithms` and put the existing dirs in reference_algorithms into `reference_algorithms/development_algorithms`. Happy to keep iterating on ideas.

fsschneider:
Looks good to me.
I think it is clear (enough) for submitters now to find what they are looking for.

runame:
Makes sense, let's just keep it like this for now.

runame (approved):

PR #180, opened 2022-10-12
Add `use_wandb` flag + refactor use of workload class attributes

runame:
This PR
- adds a `use_wandb` flag, so having the package in the env does not force you to use it,
- fixes some minor things in the submission runner and removes a redundant device transfer in the OGBG workload,
- and refactors the use of workload class attributes (only use class attributes for constants which should be the same for all class instances).


pomonam (approved):
Looks good to me. The refactor of workload class attributes makes sense to me.

znado (approved):

PR #181, opened 2022-10-12
Checkpoint utilities to submission_runner.py

pomonam:
This PR implements checkpointing for preemption. 

Notably, this PR does the following:
- Add `checkpoint_utils` that contains various functions for checkpointing.
- Merge bookkeeping variables (e.g., `goal_reached`) to a dictionary so that it is easy to save and load from a checkpoint.
- Override hyperparameters if `hparams.json` already exists. 
- Update PyTorch version to 1.12.1 by default due to an issue in https://github.com/pytorch/pytorch/issues/80809#issuecomment-1175211598. PyTorch 1.12.0 has a bug in loading Adam/AdamW optimizer checkpoints. 

Todos:
- [x]  The code was tested on a single-gpu and it works fine. I will also evaluate DDP settings to make sure it works as expected. 
- [x] There is some overlap with the PR #180. I can merge two branches once #180  gets committed.


algorithmic_efficiency/checkpoint_utils.py
```diff
@@ -0,0 +1,162 @@
...
+                    preemption_count: int,
+                    checkpoint_dir: str) -> None:
+  if RANK == 0:
+    if framework == 'jax':
+      model_params = jax.device_get(jax_utils.unreplicate(model_params))
+      opt_state, _ = optimizer_state
+      opt_state = jax.device_get(jax_utils.unreplicate(opt_state))
+      model_state = jax.device_get(jax_utils.unreplicate(model_state))
+    else:
+      model_params = model_params.state_dict()
```
runame, line 134: To also support DP instead of only DDP, I think we can just add these two lines above this one:
```python
if isinstance(model_params, torch.nn.DataParallel):
  model_params = model_params.module
```

algorithmic_efficiency/checkpoint_utils.py
```diff
@@ -0,0 +1,162 @@
...
+        ])
+    checkpoint_state['optimizer_state'] = (checkpoint_state['optimizer_state'],
+                                           opt_update_fn)
+    checkpoint_state['eval_results'] = [
+        (value, key) for key, value in latest_ckpt['eval_results'].items()
+    ]
+
+  else:
+    checkpoint_state = latest_ckpt
+    model_params.load_state_dict(checkpoint_state['model_params'])
```
runame, line 90: Do we need a `map_location` argument here (even when the state dict was saved from GPU and ist loaded to GPU, it might be necessary or at least safer to specify the correct device when using DDP)?

algorithmic_efficiency/checkpoint_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    checkpoint_state['model_params'] = model_params
+    for key in optimizer_state.keys():
+      optimizer_state[key].load_state_dict(
```
runame, line 93: See question above.

runame (approved):
Nice, this will be very useful!

algorithmic_efficiency/checkpoint_utils.py
```diff
@@ -0,0 +1,162 @@
...
+  if framework == 'jax':
+    latest_ckpt = flax_checkpoints.restore_checkpoint(
+        checkpoint_dir, target=checkpoint_state)
+    save_path = os.path.join(checkpoint_dir,
+                             'checkpoint_' + str(latest_ckpt['global_step']))
+  else:
+    latest_ckpt = checkpoint_state
+    save_path = os.path.join(checkpoint_dir, 'checkpoint')
+    if os.path.exists(save_path):
+      latest_ckpt = torch.load(save_path, map_location=DEVICE)
```
pomonam, line 58: @runame I perform the `map_location` when loading the checkpoint here.

algorithmic_efficiency/checkpoint_utils.py
pomonam: Please see my below comment, regarding the device map location.

algorithmic_efficiency/checkpoint_utils.py
runame: Ah perfect, thanks.

algorithmic_efficiency/checkpoint_utils.py
pomonam: Thank you! Both with and without did not cause problems, but I did find some threads recommending to `.module` for DataParallel models. I updated the code to do this both when saving and loading.

runame: Maybe it works if you also load the state dict to a model wrapped with DataParallel, but if not it would fail?

pomonam: Ah, that is a good point. Yes, I think that might be the case, since `nn.DataParallel` parent is `nn.Module`, so it should have functionality the load and save state_dict.

Issue #186, opened 2022-10-13
imagenet pytorch workloads break during eval on test set


priyakasimbeg:
imagenet pytorch ViT and Resnet workloads break during eval on imagenet_v2 test split.

## Description
Traceback:
```
I1013 18:53:17.879712 140653770622784 submission_runner.py:243] Initializing metrics bundle.
I1013 18:53:17.879878 140653770622784 submission_runner.py:257] Starting training loop.
I1013 18:53:47.672241 140653770622784 spec.py:288] Evaluating on the training split.
I1013 18:55:05.126325 140653770622784 spec.py:300] Evaluating on the validation split.
I1013 18:56:16.793108 140653770622784 spec.py:315] Evaluating on the test split.
I1013 18:56:16.796242 140653770622784 dataset_info.py:358] Load dataset info from /home/kasimbeg/imagenet_v2/matched-frequency/3.0.0
I1013 18:56:16.798682 140653770622784 dataset_builder.py:351] Reusing dataset imagenet_v2 (/home/kasimbeg/imagenet_v2/matched-frequency/3.0.0)
I1013 18:56:16.798848 140653770622784 logging_logger.py:36] Constructing tf.data.Dataset imagenet_v2 for split test, from /home/kasimbeg/imagenet_v2/matched-frequency/3.0.0
Traceback (most recent call last):
  File "submission_runner.py", line 495, in <module>
    app.run(main)
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "submission_runner.py", line 479, in main
    FLAGS.tokenizer_vocab_path)
  File "submission_runner.py", line 411, in score_submission_on_workload
    tokenizer_vocab_path)
  File "submission_runner.py", line 307, in train_once
    global_step)
  File "/home/kasimbeg/algorithmic-efficiency/algorithmic_efficiency/spec.py", line 324, in eval_model
    global_step=global_step)
  File "/home/kasimbeg/algorithmic-efficiency/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py", line 249, in _eval_model_on_split
    batch = next(self._eval_iters[split])
[...]

PR #187, opened 2022-10-13
fix axes mismatch bug in case DDP is False

priyakasimbeg:
Fix for axes bug in imagenet_v2 data pipeline. IF USE_PYTORCH_DDP is false we should not attempt to index data w `v[RANK]`. This should resolve https://github.com/mlcommons/algorithmic-efficiency/issues/186.


runame (approved):
LGTM

Issue #189, opened 2022-10-14
Check initializations for JAX and PyTorch are the same.


pomonam:
We need to make sure the initialization methods for JAX and PyTorch are the same. 

Moreover, making a unit test to check the outputs are almost the same when the model params are the same for JAX and PyTorch and we are using the same mini-batch. We may need to hard-code the correspondence between Jax `param_key` and PyTorch `param_key`, but it would be useful as it shows that the same operations are being performed.

chandramouli-sastry:
i'll work on testing the conformer model!

pomonam:
Note that the initialization schemes are now all unified except for librispeech workloads.

pomonam:
Closing with #195.

Issue #184, opened 2022-10-13
imagenet workloads not converging at expected rate


priyakasimbeg:
The imagenet validation error rate vs step does not line up with reference runs for the resnet and vit workloads in both jax and pytorch.

## Description
Imagenet vit: 
- expected: 140K steps val accuracy 0.77171
- actual (jax): 187K steps val accuracy 0.5498
- actual (pytorch): 128K steps val accuracy 0.2339

Imagenet Resnet:
- expected: 140K steps val accuracy 0.77165
- actual (jax): 190K steps val accuracy 0.765680
- actual (pytorch):  pending 

## Steps to Reproduce
Run commands for Imagenet vit and Imagenet Resnet in [target setting runs](https://github.com/mlcommons/algorithmic-efficiency/blob/e6a5d0f8f4c8687aeaad0a738b9722dc8d1e5806/reference_algorithms/target_setting_algorithms/README.md).

## Source or Possible Fix
One possible source is difference in hparams, but I checked that the hparams ([imagenet resnet](reference_algorithms/target_setting_algorithms/imagenet_resnet/tuning_search_space.json) and [imagenet vit](reference_algorithms/target_setting_algorithms/imagenet_vit/tuning_search_space.json)) match the [reference hparams](https://docs.google.com/spreadsheets/d/1-OZZ_mjtevzFoNIJA5x7iJtd0egnV3BOqf5Ft1anUW8/htmlview?resourcekey=0-gj4v0Hj20h1rDILq4FZ9Gw#gid=538625624).

znado:
We need to implement RandAug.

pomonam:
Just for the update, the `Imagenet Resnet` workload should be ready for a re-run. We were applying `log_softmax` twice and is surprising that it could even get an accuracy of 76.5. 

In terms of the ViT, I will implement RandAug soon and then we should be ready for a re-run.

priyakasimbeg:
Hi Juan thanks for all the hard work in resolving all the differences. I reran the Imagenet Resnet workloads and the jax workload got VERY close to the target after about 140K steps (which is what the reference internal run got and what the step hint for the learning rate schedule was set to). The target is 0.77185 and the actual max value during evals was 0.77175. We think it missed it by random chance.

![image (2)](https://user-images.githubusercontent.com/12614254/197259343-056fb57e-7a86-4e2f-8109-f4e328852b46.png)

[...]

PR #193, opened 2022-10-18
ViT architecture updates

pomonam:
This PR attempts to address #184 and #189.

To match the configuration from target-setting runs:
1. Change architecture configuration to `S/16` from `B/32`.
2. Use identical parameter initialization methods for both Jax and PyTorch workloads.
3. Clean up the code by removing any intermediate unused variables.
4. Add gradient clipping for PyTorch and use gradient clipping of 1.0 in the target-setting run.
5. Fix gradient clipping for Jax (I suspect there is a missing `sqrt` if I am not mistaken).

However, there are still some differences:
1. The way we are applying weight decay (e.g., apply to tensors with above or equal to 2 dimensions vs. apply to all tensors) -- which can change the optimal LR and WD.
2. Data augmentation (e.g., using RandAugment) -- which can change the optimal LR due to the implicit regularization effect.

### DIFF
--- a/algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/models.py
+++ b/algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/models.py
@@ -9,7 +9,6 @@ from typing import Optional, Sequence, Union
 
 from flax import linen as nn
 import jax.numpy as jnp
-import numpy as np
 
 from algorithmic_efficiency import spec
 
--- a/algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/models.py
+++ b/algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/models.py
@@ -11,6 +11,8 @@ from flax import linen as nn
 import jax.numpy as jnp
 import numpy as np
 
+from algorithmic_efficiency import spec
+
 
 def posemb_sincos_2d(h, w, width, temperature=10_000., dtype=jnp.float32):
   """Follows the MoCo v3 logic."""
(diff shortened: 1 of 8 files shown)


algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/models.py
```diff
@@ -95,68 +84,49 @@ class Encoder(nn.Module):
...
   """ViT model."""
 
-  num_classes: int
+  num_classes: int = 1000
   patch_size: Sequence[int] = (16, 16)
   width: int = 768
   depth: int = 12
   mlp_dim: Optional[int] = None  # Defaults to 4x input dim
   num_heads: int = 12
-  posemb: str = 'sincos2d'  # Can also be "learn"
```
runame, line 147: General question: should we remove the code for all options which we don't use, to simplify the code?

algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/models.py
```diff
@@ -95,68 +84,49 @@ class Encoder(nn.Module):
...
+  posemb: str = 'sincos2d'
   rep_size: Union[int, bool] = True
   dropout_rate: Optional[float] = 0.0  # If None, defaults to 0.0.
-  pool_type: str = 'gap'  # Can also be 'map' or 'tok'
+  pool_type: str = 'gap'  # Can also be 'tok'
```
runame, line 111: Same question as above.

algorithmic_efficiency/workloads/imagenet_vit/imagenet_pytorch/models.py
```diff
@@ -218,9 +199,9 @@ def __init__(
       depth: int = 12,
       mlp_dim: Optional[int] = None,  # Defaults to 4x input dim
       num_heads: int = 12,
-      posemb: str = 'sincos2d',  # Can also be 'learn'
+      posemb: str = 'sincos2d',
```
runame, line 202: Same question as above.

algorithmic_efficiency/workloads/imagenet_vit/imagenet_pytorch/models.py
```diff
@@ -218,9 +199,9 @@ def __init__(
...
       rep_size: Union[int, bool] = True,
-      dropout: float = 0.0,
+      dropout_rate: float = 0.0,
       pool_type: str = 'gap',  # Can also be 'tok'
```
runame, line 205: Same question as above.

runame (approved):
LGTM

algorithmic_efficiency/workloads/imagenet_vit/imagenet_jax/models.py
pomonam: Yes, we can remove the code for all options for simplicity. I will work on it.

runame: Thanks!

znado:
Regarding your question about which tensors to apply WD to, I think it's OK to apply it to all because [here](https://github.com/google/init2winit/blob/master/init2winit/trainer_lib/trainer.py#L110) in init2winit we apply it to all (the rank threshold is 0 which is all tensors), and it looks like PyTorch also applies it to all tensors. The config for ViT is confusing because in init2winit we support both L2 regularization and WD, and so the `"l2_decay_rank_threshold": 2,` is ignored because we have `"l2_decay_factor": null,`, and then WD is applied because we have `"weight_decay": 0.026595371563300656`.

Regarding randaug, that is a real diff that we should implement, I commented on https://github.com/mlcommons/algorithmic-efficiency/issues/184. Can you make a follow-up PR that adds it? I'm not sure the best place to grab a PyTorch implementation, but you can use the tf.data one from init2winit [here](https://github.com/google/init2winit/blob/master/init2winit/dataset_lib/imagenet_dataset.py#L264).

znado:
Thanks so much for hunting down all these diffs. These were major changes that we needed to resolve, and finding them requires talented ML and engineering skills while carefully reading the code.

PR #194, opened 2022-10-18
ImageNet ResNet initialization to match with Jax

pomonam:
This PR modifies the Imagenet-Resnet workload so that the initialization scheme between Jax and Pytorch matches.

### DIFF
--- a/algorithmic_efficiency/init_utils.py
+++ b/algorithmic_efficiency/init_utils.py
@@ -1,9 +1,15 @@
+"""Utilities for initializing parameters.
+
+Note: Code adapted from
+https://github.com/google/jax/blob/main/jax/_src/nn/initializers.py
+"""
 import math
 
 from torch import nn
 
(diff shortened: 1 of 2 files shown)


algorithmic_efficiency/init_utils.py
```diff
@@ -0,0 +1,11 @@
+import math
+
+from torch import nn
+
+
```
runame, line 5: Maybe it makes sense to add a comment with the [reference](https://jax.readthedocs.io/en/latest/_autosummary/jax.nn.initializers.lecun_normal.html) to the corresponding Jax initialization function here?

runame (approved):
LGTM

algorithmic_efficiency/init_utils.py
pomonam: That's a great point. I will add that.

PR #195, opened 2022-10-18
Initalization Fixes

pomonam:
This PR addresses issues that appeared at #189. 

After this PR, the initialization scheme for Jax and Pytorch should be identical except for librispeech workload. Librispeech PyTorch workload might require a larger change as it is difficult to set the initialization for LazyLinear. We might need to define our own version of LazyLinear and use Xavier initialization.


runame (approved):
LGTM

PR #199, opened 2022-10-18
More ResNet Diff Checking

pomonam:
I believe that we can run the target-setting for ImageNet ResNet after this PR. Basically, this PR attempts to close the gap with the code from target-setting runs.

- Fix final dense layer initialization to normal with `stddev=1e-2` (for both Jax and PyTorch).
- Remove `log_softmax` inside ResNet and let it apply `log_softmax` in the loss function.


znado (approved):
Nice catches, LGTM!

PR #205, opened 2022-10-19
timing fix

znado:
Setting `last_eval_time` to be the current time at the end of an eval, after the eval, checkpointing, and logging has finished. Previously we were setting `last_eval_time` to be the time the eval started at, which meant that the "time since last eval" calculations were counting the time spent in eval/checkpointing/logging as part of trianing.

### DIFF
--- a/algorithmic_efficiency/checkpoint_utils.py
+++ b/algorithmic_efficiency/checkpoint_utils.py
@@ -160,7 +160,8 @@ def save_checkpoint(framework: str,
     else:
       save_path = os.path.join(checkpoint_dir, 'checkpoint')
       torch.save(checkpoint_state, save_path)
-    logging.info('Saved checkpoint to %s', save_path)
 
   if USE_PYTORCH_DDP:
     dist.barrier()
+  if RANK == 0:
+    logging.info('Saved checkpoint to %s', save_path)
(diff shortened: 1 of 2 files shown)


algorithmic_efficiency/checkpoint_utils.py
```diff
@@ -160,7 +160,8 @@ def save_checkpoint(framework: str,
     else:
       save_path = os.path.join(checkpoint_dir, 'checkpoint')
       torch.save(checkpoint_state, save_path)
-    logging.info('Saved checkpoint to %s', save_path)
 
   if USE_PYTORCH_DDP:
     dist.barrier()
+  if RANK == 0:
```
runame, line 166: `logging.info()` will not do anything in any RANK != 0 anyway, see [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/pytorch_utils.py#L41-L46), so this change is not necessary I think.

pomonam (approved):
LGTM!

runame (approved):
Have to check the whole timing flow at a later point, but for now this makes sense to me.

algorithmic_efficiency/checkpoint_utils.py
znado: I originally didn't have the check, but then `save_path` isn't defined so I had to add it.

runame: Ah right, I was too tired to catch that.

PR #206, opened 2022-10-19
Additional metrics logging across workloads - batch_size, num_examples, grad_norm, loss

sourabh2k15:


reference_algorithms/target_setting_algorithms/jax_submission_base.py
```diff
@@ -91,9 +90,9 @@ def update_params(workload: spec.Workload,
...
   else:
     grad_clip = None
-  new_optimizer_state, new_params, new_model_state = pmapped_train_step(
+  new_optimizer_state, new_params, new_model_state, loss, grad_norm = pmapped_train_step( # pylint: disable=line-too-long
       workload, opt_update_fn, model_state, optimizer_state,
       current_param_container, batch, per_device_rngs, grad_clip,
       label_smoothing)
 
-  return (new_optimizer_state, opt_update_fn), new_params, new_model_state
+  return (new_optimizer_state, opt_update_fn), new_params, new_model_state, loss.mean(), grad_norm.mean()  # pylint: disable=line-too-long
```
znado, line 98: the API for a submission's `update_params` function says we have to return `new_optimizer_state, new_params, new_model_state`, so we can't return these here.

but what we can do instead is replace this line's change with:
```python
      # Log training metrics - loss, grad_norm, batch_size.
      if global_step <= 100 or global_step % 500 == 0:
          workload.metrics_logger.append_scalar_metrics(
              {
                  'loss': loss,
                  'grad_norm': grad_norm,
              },
              global_step,
              preemption_count)
        logging.info('%d) loss = %0.3f, grad_norm = %0.3f',
                     global_step,
                     loss,
                     grad_norm)
      return (new_optimizer_state, opt_update_fn), new_params, new_model_state
```

and then undo the changes in submission_runner.py. This doesn't log `global_batch_size` but I think we can just add that to the json files we log at the beginning if we really need it.

znado (changes requested):

znado (approved):

PR #208, opened 2022-10-20
RandAug Implementation for ViT

pomonam:
This PR implements RandAug for ViT. 

- Note that I don't have an environment set up at the moment and I need to run the code to make sure that nothing breaks.
- PyTorch RandAugment does not include `Cutout` or `AutoContrast` at the moment and the behaviour might be slightly different from the one in JAX. This is a quick patch, but we ideally need to (1) unify the input pipeline or (2) port RandAug operations in Jax to PyTorch.


znado (approved):
LGTM, merging for now, but as discussed before it looks like we also need to add the following to PyTorch to make it equivalent to the tf version:
Invert
SolarizeAdd
Cutout

PR #220, opened 2022-10-24
RandAug Implementation for ViT

pomonam:
This PR implements RandAug for both Jax and PyTorch workloads. Previous bugs are fixed and PyTorch RandAugment code is manually modified so that it matches that of Jax. 

Note: I also noticed one additional difference with the target-setting run; the target-setting run uses an 'inception' crop, whereas we are using a 'random' crop.


znado (approved):
LGTM!

PR #221, opened 2022-10-24
Max global steps

priyakasimbeg:
Add flag for max_global_steps.

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -302,7 +302,7 @@ def train_once(
       train_state['training_complete'] = True
     global_step += 1
     if (max_global_steps is not None) and (global_step == max_global_steps):
-      training_state['training_complete'] = True
+      train_state['training_complete'] = True
     if USE_PYTORCH_DDP:
       # Make sure all processes run eval after the same step when using DDP.
       dist.barrier()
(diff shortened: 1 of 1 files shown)


submission_runner.py
```diff
@@ -300,6 +301,8 @@ def train_once(
     except spec.TrainingCompleteError:
       train_state['training_complete'] = True
     global_step += 1
+    if global_step == max_global_steps:
```
znado, line 304: do we also need a check for `if max_global_steps is not None and global_step == max_global_steps:` ?

submission_runner.py
```diff
@@ -129,7 +129,7 @@
                      False,
                      'Whether to use Weights & Biases logging.')
 flags.DEFINE_boolean('profile', False, 'Whether to produce profiling output.')
-
+flags.DEFINE_integer('max_global_steps', None, 'Maximum number of update steps')
```
znado, line 132: nit: add `.` to end of help string.

znado (approved):
LGTM just minor fix

submission_runner.py
priyakasimbeg: If max_global_steps is None then global_step == max_global_steps will be False. So I don't think we need to add an extra check.

Issue #198, opened 2022-10-18
ogbg returns NaN error 


priyakasimbeg:
The ogbg jax workload returned `ValueError: Input contains NaN, infinity or a value too large for dtype('float32')` after about 72 steps. 
                                                                                          
## Description
Traceback:
```
I1018 21:44:36.960185 140585414121280 checkpoints.py:356] Saving checkpoint at step: 72                                                                                            
I1018 21:44:37.026029 140585414121280 checkpoints.py:317] Saved checkpoint at /home/kasimbeg/target_setting/ogbg_jax/trial_1/checkpoint_72                                         
I1018 21:44:37.026265 140585414121280 checkpoints.py:344] Removing checkpoint at /home/kasimbeg/target_setting/ogbg_jax/trial_1/checkpoint_2                                       
I1018 21:44:37.029724 140585414121280 checkpoint_utils.py:163] Saved checkpoint to /home/kasimbeg/target_setting/ogbg_jax/trial_1/checkpoint_72                                    
I1018 21:46:39.350129 140585414121280 spec.py:288] Evaluating on the training split.                                                                                               
Traceback (most recent call last):                                                                                                                                                 
  File "submission_runner.py", line 545, in <module>                                                                                                                               
    app.run(main)                                                                                                                                                                  
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/absl/app.py", line 312, in run                                                                       
    _run_main(main, args)                                                                                                                                                          
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/absl/app.py", line 258, in _run_main                                                                 
    sys.exit(main(argv))                                                                                                                                                           
  File "submission_runner.py", line 529, in main                                                                                                                                   
    FLAGS.tokenizer_vocab_path)                                                                                                                                                    
  File "submission_runner.py", line 463, in score_submission_on_workload                                                                                                           
    tokenizer_vocab_path)                                                                                                                                                          
  File "submission_runner.py", line 323, in train_once                                                                                                                             
    global_step)                                                                                                                                                                   
  File "/home/kasimbeg/algorithmic-efficiency/algorithmic_efficiency/spec.py", line 297, in eval_model                                                                             
    global_step=global_step)                                                                                                                                                       
  File "/home/kasimbeg/algorithmic-efficiency/algorithmic_efficiency/workloads/ogbg/workload.py", line 138, in _eval_model_on_split                                                
[...]

PR #228, opened 2022-10-26
Ogbg fix nans

priyakasimbeg:
Verified that this should fix  https://github.com/mlcommons/algorithmic-efficiency/issues/198.


runame (approved):
LGTM! I guess we somehow have to communicate to users that some losses can have NaN values in them, so they can use `nanmean` instead of `mean`.

PR #233, opened 2022-10-27
Pad remainder batch during eval

### LINKED ISSUE #119: Check for correctly divisible batch size
## Description

We should probably throw an informative error if
1) the training batch size is not divisible by the number of GPUs which are used,
2) the evaluation batch size is chosen such that the size of the remainder is not divisible by the number of GPUs which are used.

runame:
This PR fixes #119.

I noticed the following things, were they done intentionally? (I already addressed all of them but the last point.)
- The ImageNet V2 test set was already padded, but the `weights` were not used anywhere in the workloads.
- In the FastMRI workloads the weights are only used for the denominator, but not in the calculation of the SSIM score and loss.
- In the Criteo1TB workloads the weights seemed to be always set to all ones, i.e. they didn't make any difference.
- The librispeech workloads already implement padding, but do not create/use a mask.

**TODO:**
- [ ] The change for Criteo1TB is implemented but not tested, because I didn't have the data on my setup. Can I easily access the data from GCP? Then I will test the changes there.
- [ ] Use the new padding utils for librispeech as well and fix the lack of the mask.


znado:
> * The ImageNet V2 test set was already padded, but the `weights` were not used anywhere in the workloads.
We should definitely use them.

> * In the FastMRI workloads the weights are only used for the denominator, but not in the calculation of the SSIM score and loss.
That sounds like a bug?

> * In the Criteo1TB workloads the weights seemed to be always set to all ones, i.e. they didn't make any difference.
We should make the final partial batch padded, I'm not sure how this was working before.

> * The librispeech workloads already implement padding, but do not create/use a mask.
For the speech workloads we also have target padding, so those were handled differently. Not sure if that also handled the final partial batch, but if not we should fix that. However I could also see a reasonable choice being that instead of having the inputs be a tuple of (inputs, input_paddings) and targets being (targets, target_paddings) and mask being None, we would have inputs be inputs, targets be targets, and mask being a tuple of (input_paddings, target_paddings). I'll defer to @sourabh2k15 for which he prefers though.

> * [ ]  The change for Criteo1TB is implemented but not tested, because I didn't have the data on my setup. Can I easily access the data from GCP? Then I will test the changes there.
Yep, you can copy them via `gsutil -m cp -r [/home/znado](gs://mlcommons-data/criteo/criteo_parts/ <local_dir>`.

runame:
> We should make the final partial batch padded, I'm not sure how this was working before.

Yeah I have done that. I should have included a link, [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/criteo1tb/input_pipeline.py#L49) the weights are created, but there is no padding used and the weights are only used to calculate the denominator in the workloads. I assume this is just a bug, same as for ImageNet and FastMRI.

> For the speech workloads we also have target padding, so those were handled differently. (...)

The final partial batch is padded, i.e. all four tensors `inputs`, `input_paddings`, `targets`, and `target_paddings` are padded. However, in contrast to everywhere else, they are padded with `1`s instead of `0`s and no weights are created. @sourabh2k15 Can you confirm that this leads to the desired behavior in the CTC loss and the other metrics?

> Yep, you can copy them via `gsutil -m cp -r gs://mlcommons-data/criteo/criteo_parts/ <local_dir>`.

Nice, thanks!

algorithmic_efficiency/data_utils.py
```diff
@@ -7,28 +9,63 @@
 from torch.utils.data import DistributedSampler
 from torch.utils.data import Sampler
 
+from algorithmic_efficiency import spec
+
 
-def shard_numpy_ds(xs):
+def shard_numpy_ds(
```
runame, line 15: Maybe we should rename this function to something like `shard_and_maybe_pad_ds()`? I guess the full descriptive name `convert_to_numpy_and_shard_and_maybe_pad_ds()` is a bit long haha

znado (approved):
Looks fantastic, thanks so much for going through all the workloads and doing this, it's very important work that has to be carefully executed!

znado:
I'm going to go ahead and merge this so other people can use it ASAP, if you want to rename `shard_numpy_ds` submit a follow-up PR and I'll review it quickly.

algorithmic_efficiency/data_utils.py
```diff
@@ -7,28 +9,76 @@
...
+    # We need a 2d mask for WMT.
+    mask_shape = targets_shape if len(targets_shape) < 3 else targets_shape[0]
+    xs['weights'] = create_mask(mask_shape, pad_size, 'jax')
 
   def _prepare(x):
     # Use _numpy() for zero-copy conversion between TF and NumPy.
     if not isinstance(x, np.ndarray):
       x = x._numpy()  # pylint: disable=protected-access
 
+    # Pad if remainder_size != 0 and x is not value for key 'weights'.
```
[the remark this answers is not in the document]
runame: I already changed this comment before your review ;) Before I was creating the mask and had to choose the condition here to not pad the mask, but now I just create the first part of the mask and pad it with the rest of the batch.

algorithmic_efficiency/data_utils.py
runame: Yeah I like that suggestion, I'll include this change in a follow-up PR.

PR #234, opened 2022-10-31
Fix usage of weights in WMT workloads + clean-up

### LINKED ISSUE #227: WMT Pytorch loss fluctuates over time
For the WMT Pytorch workload the loss and other metrics decrease, increase and decrease again over time. 

## Description
image

## Steps to Reproduce
On kasimbeg-6 on branch [sacrebleu](https://github.com/priyakasimbeg/algorithmic-efficiency/tree/sacrebleu) run 
```
torchrun --redirects 1:0,2:0,3:0,4:0,5:0,6:0,7:0 --standalone --nnodes=1 --nproc_per_node=8 submission_runner.py \
    --framework=pytorch \
    --data_dir=/home/kasimbeg/tensorflow_datasets \
    --experiment_dir=/home/kasimbeg \
    --experiment_name=target_setting \
    --workload=wmt \
    --submission_path=reference_algorithms/target_setting_algorithms/pytorch_adamw.py \
    --tuning_search_space=reference_algorithms/target_setting_algorithms/wmt/tuning_search_space.json > results_wmt_pytorch.txt 2>&1
```

## Source or Possible Fix

Currently investigating init, optimizers etc.

runame:
This PR is a follow-up to #233. It 
- renames `shard_numpy_ds` to `shard_and_maybe_pad_np`,
- refactors padding in librispeech workloads to use `shard_and_maybe_pad_np`,
- adds more type hints to the WMT workloads,
- refactors the use of weights in the WMT workloads, which fixes #227 (at least it should, I did not train long enough to check),
- and changes the `loss_fn` API to return a tuple `(correct averaged loss, per-example losses)` instead of just a tensor `per_example_losses`.

### DIFF
--- a/algorithmic_efficiency/workloads/ogbg/workload.py
+++ b/algorithmic_efficiency/workloads/ogbg/workload.py
@@ -1,6 +1,6 @@
 import itertools
 import math
-from typing import Dict
+from typing import Dict, Optional, Tuple
 
 from absl import flags
 import jax
@@ -81,17 +81,25 @@ class BaseOgbgWorkload(spec.Workload):
 
   # Does NOT apply regularization, which is left to the submitter to do in
   # `update_params`.
-  def loss_fn(self,
-              label_batch: spec.Tensor,
-              logits_batch: spec.Tensor,
-              mask_batch: spec.Tensor,
-              label_smoothing: float = 0.0) -> spec.Tensor:  # differentiable
+  def loss_fn(
+      self,
+      label_batch: spec.Tensor,
+      logits_batch: spec.Tensor,
+      mask_batch: Optional[spec.Tensor] = None,
+      label_smoothing: float = 0.0
+  ) -> Tuple[spec.Tensor, spec.Tensor]:  # differentiable
+    """Return (correct scalar average loss, 1-d array of per-example losses)."""
     per_example_losses = self._binary_cross_entropy_with_mask(
         labels=label_batch,
         logits=logits_batch,
         mask=mask_batch,
         label_smoothing=label_smoothing)
-    return per_example_losses
+    if mask_batch is not None:
+      n_valid_examples = mask_batch.sum()
+    else:
+      n_valid_examples = len(per_example_losses)
+    summed_loss = per_example_losses.sum()
+    return summed_loss / n_valid_examples, per_example_losses
 
   @property
   def step_hint(self) -> int:
--- a/algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
@@ -25,11 +25,12 @@ USE_PYTORCH_DDP, RANK, DEVICE, N_GPUS = pytorch_utils.pytorch_setup()
 class WmtWorkload(BaseWmtWorkload):
   """WMT PyTorch workload."""
 
-  def compute_weighted_cross_entropy(self,
-                                     logits,
-                                     targets,
-                                     weights=None,
-                                     label_smoothing=0.1):
+  def compute_weighted_cross_entropy(
+      self,
+      logits: spec.Tensor,
+      targets: spec.Tensor,
+      weights: Optional[spec.Tensor] = None,
+      label_smoothing: float = 0.1) -> spec.Tensor:
     """Compute weighted cross entropy and entropy for log probs and targets.
 
     Args:
@@ -60,7 +61,12 @@ class WmtWorkload(BaseWmtWorkload):
   # Primary eval / decode step functions.
   # ----------------------------------------------------------------------------
   @torch.no_grad()
-  def predict_step(self, inputs, params, eos_id, max_decode_len, beam_size=4):
+  def predict_step(self,
+                   inputs: spec.Tensor,
+                   params: spec.ParameterContainer,
+                   eos_id: int,
+                   max_decode_len: int,
+                   beam_size: int = 4) -> spec.Tensor:
     """Predict translation with fast decoding beam search on a batch."""
     params = params.module if isinstance(params, (DP, DDP)) else params
     params.eval()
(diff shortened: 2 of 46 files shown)


algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
```diff
@@ -43,8 +44,8 @@ def compute_weighted_cross_entropy(self,
       Tuple of loss for every example and batch normalizing factor.
```
znado, line 44: This docstring is no longer accurate, right? It should just be per-example losses?

znado (approved):
LGTM, just one question about NaNs as padding

algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
runame: Nice catch, fixed.

algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
```diff
@@ -289,18 +298,26 @@ def _build_input_queue(self,
...
         rng=None,
         update_batch_norm=False)
-    return self.compute_summed_metrics(logits, targets, weights)
+    loss = self.compute_weighted_cross_entropy(logits, targets, weights, 0.0)
+    mask = torch.where(targets > 0, 1, 0)
+    if weights is not None:
+      mask = torch.logical_and(weights, mask)
+    acc_sum, weight_sum = self.compute_weighted_accuracy(logits, targets, mask)
+    return {
+        'loss': torch.nansum(loss),
```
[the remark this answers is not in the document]
runame: Ah nice catch, you are completely right! This also affects the change in the OGBG workload (see [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/ogbg/ogbg_pytorch/workload.py#L200)). The reason why I followed the fix, i.e. padding with nans, is that otherwise we require the user to not use the mean, but to sum the losses and divide by `weights.sum()` (assuming we add the weights in the training `batch`, otherwise we would require the user to know which weights to use). See the discussion in #227. 

Since the instruction on how to average the loss should be workload agnostic, we would have to provide very specific instructions, like saying the users have to set `denominator = batch.get('weights').sum() if 'weights' in batch else len(losses)` or something like this. If the user doesn't follow the instruction there won't be any obvious error; hence, we thought padding with nans is better. But I guess we don't have any other option than to provide explicit instructions on how to average the loss?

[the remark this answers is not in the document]
runame: I agree, I think changing the `loss_fn` API to return a tuple is probably best. We could either go with `(safely averaged loss, per-example losses)` or with `(per-examples losses, correct denominator)`. The advantage of the first version is that many (most?) people probably just want the averaged loss anyway, so they can just get the first output element and don't have to do anything more with it. Also, in the case where the user wants to use the per-example losses, the correct denominator can be derived from this case as well, so the information content of both versions is the same.

If you also think that the first version is best, I will implement it for all workloads in this PR.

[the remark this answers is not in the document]
runame: I went ahead and just implemented this. Let me know if you prefer a different solution.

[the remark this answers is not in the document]
znado: yep I prefer the first one too, thanks for implementing :)

algorithmic_efficiency/workloads/ogbg/workload.py
```diff
@@ -81,17 +81,25 @@ def _build_input_queue(self,
...
+  ) -> Tuple[spec.Tensor, spec.Tensor]:  # differentiable
+    """Return (correct scalar average loss, 1-d array of per-example losses)."""
     per_example_losses = self._binary_cross_entropy_with_mask(
         labels=label_batch,
         logits=logits_batch,
         mask=mask_batch,
         label_smoothing=label_smoothing)
-    return per_example_losses
+    if mask_batch is not None:
+      n_valid_examples = mask_batch.sum()
```
znado, line 98: for OGBG, the masking is really weird (annoying), where it is actually 2-D because it has per-example and _per-class_ masking, so I think this has to be "number of examples that have a mask that is _all_ zeros" instead of just a sum

znado (changes requested):

algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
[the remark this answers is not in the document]
znado: +1, implement it in a follow-up PR

algorithmic_efficiency/workloads/ogbg/workload.py
runame: Hm but this is equivalent to how it was before (see [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/reference_algorithms/development_algorithms/ogbg/ogbg_pytorch/submission.py#L60-L62)), unless I'm missing something. Or was it wrong before already?

runame: Hm but this is equivalent to how it was before (see [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/reference_algorithms/development_algorithms/ogbg/ogbg_pytorch/submission.py#L60-L62)), unless I'm missing something. Or was it wrong before already?

znado: ok, I think I confused myslef, and this is the way that we do the padding in init2winit. I think you're right because this is basically doing a per-class sigmoid_binary_cross_entropy calculation, and we want to mask out some classes too, so it's valid to sum across the whole mask. sorry for the confusion!!

znado (approved):

PR #235, opened 2022-11-01
Dataset setup

priyakasimbeg:
Add dataset_setup scripts for downloading, and organizing datasets.
Originally forked from znado/algorithmic-efficiency/ds-setup.


znado (approved):
LGTM, some things are still a WIP but we will keep iterating!

PR #239, opened 2022-11-02
Fix `loss_fn` in librispeech workloads

runame:
This PR is a follow-up to #234 and fixes the `loss_fn` in the librispeech workloads to adhere to the new spec. Moreover, I noticed that in the PyTorch workloads the loss was synced via `all_reduce` in `loss_fn`. This is a bug, since it introduces unnecessary communication overhead when using PyTorch DDP (and can maybe even lead to wrong behavior, not sure about that).

### DIFF
--- a/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
+++ b/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
@@ -93,23 +93,33 @@ class LibriSpeechConformerWorkload(workload.BaseLibrispeechWorkload):
           mutable=False)
       return (logits, logit_paddings), model_state
 
-  def loss_fn(self,
-              label_batch: Tuple[spec.Tensor, spec.Tensor],
-              logits_batch: Tuple[spec.Tensor, spec.Tensor],
-              mask_batch: Optional[spec.Tensor] = None,
-              label_smoothing: float = 0.0) -> spec.Tensor:  # differentiable
-    del mask_batch
+  # Does NOT apply regularization, which is left to the submitter to do in
+  # `update_params`.
+  def loss_fn(
+      self,
+      label_batch: spec.Tensor,
+      logits_batch: spec.Tensor,
+      mask_batch: Optional[spec.Tensor] = None,
+      label_smoothing: float = 0.0
+  ) -> Tuple[spec.Tensor, spec.Tensor]:  # differentiable
+    """Return (correct scalar average loss, 1-d array of per-example losses)."""
     del label_smoothing
     logits, logit_paddings = logits_batch
     targets, target_paddings = label_batch
     logprobs = nn.log_softmax(logits)
-    per_seq_loss = self.ctc_loss(logprobs,
-                                 logit_paddings,
-                                 targets,
-                                 target_paddings)
-    normalizer = jnp.sum(1 - target_paddings)
-    normalized_loss = jnp.sum(per_seq_loss) / jnp.maximum(normalizer, 1)
-    return normalized_loss
+    per_example_losses = self.ctc_loss(logprobs,
+                                       logit_paddings,
+                                       targets,
+                                       target_paddings)
+    # mask_batch is assumed to be shape [batch].
+    if mask_batch is not None:
+      per_example_losses *= mask_batch
+      mask_batch = jnp.logical_and(mask_batch, 1 - target_paddings)
+    else:
+      mask_batch = target_paddings
+    n_valid_examples = jnp.maximum(mask_batch.sum(), 1)
+    summed_loss = per_example_losses.sum()
+    return summed_loss / n_valid_examples, per_example_losses
 
   def ctc_loss(self,
                logits: spec.Tensor,
@@ -214,7 +224,8 @@ class LibriSpeechConformerWorkload(workload.BaseLibrispeechWorkload):
         update_batch_norm=False)
 
     decoded, decoded_paddings = self.greedy_decode(logits, logit_paddings)
-    normalized_loss = self.loss_fn(batch['targets'], (logits, logit_paddings))
+    normalized_loss, _ = self.loss_fn(
+        batch['targets'], (logits, logit_paddings))
 
     targets, target_paddings = batch['targets']
     return self.metrics_bundle.gather_from_model_output(
--- a/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
+++ b/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
@@ -116,7 +116,7 @@ class LibriSpeechConformerWorkload(workload.BaseLibrispeechWorkload):
       per_example_losses *= mask_batch
       mask_batch = jnp.logical_and(mask_batch, 1 - target_paddings)
     else:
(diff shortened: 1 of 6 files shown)


znado (approved):

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
```diff
@@ -93,23 +93,33 @@ def model_fn(
...
+                                       logit_paddings,
+                                       targets,
+                                       target_paddings)
+    # mask_batch is assumed to be shape [batch].
+    if mask_batch is not None:
+      per_example_losses *= mask_batch
+      mask_batch = jnp.logical_and(mask_batch, 1 - target_paddings)
+    else:
+      mask_batch = target_paddings
+    n_valid_examples = jnp.maximum(mask_batch.sum(), 1)
```
sourabh2k15, line 120: this should be `n_valid_examples  = jnp.maximum(1 - mask_batch.sum(), 1)`,  given mask_batch is target_paddings if it's None

in librispeech padding 1.0 means padded value

runame: Great catch, fixed!

sourabh2k15:
Hi Runa, I wanted to better understand the pytorch part before we merge this ... are we sure this is a bug and we need to fix it ? 

asking cause Chandramouli has a test comparing previous state of jax librispeech loss function to previous state of torch librispeech loss function 

at the very least we need to run that test and confirm it passes, I'd also request you to run the updated loss functions and make sure they get exactly same values as before

sourabh2k15:
also you could include changes from https://github.com/mlcommons/algorithmic-efficiency/pull/241 

I'll close that PR if those changes are included here , but it's absolutely needed after loss API changes , currently librispeech workloads are broken at head

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -266,15 +266,18 @@ def _eval_model_on_split(self,
...
+          decoded=decoded.cpu().numpy(),
+          decoded_paddings=decoded_paddings.cpu().numpy(),
+          targets=targets.cpu().numpy(),
+          target_paddings=target_paddings.cpu().numpy(),
           tokenizer=self.tokenizer)
-      loss = self._loss_fn((targets, target_paddings), (logits, logits_padding))
+      mean_loss, per_example_losses = self.loss_fn(
+          (targets, target_paddings), (logits, logits_padding))
+      summed_loss = per_example_losses.sum()
+      lengths = summed_loss / mean_loss
```
chandramouli-sastry, line 277: While this is correct theoretically, we might need to do a round operation to ensure that the `lengths` are whole numbers? In the eval step, we `all_reduce` the lengths and then use it for computing globally normalized ctc loss and am not totally sure if this would distort the loss-values in some way and by how much -- at this point, im okay either way :)

runame:
> The all_reduce in the loss function is not a bug, but i agree that it is a overhead. Essentially, average of normalized-ctc-losses is not equal to the globally normalized ctc-loss. On backward, all the processes will have the same gradients for all parameters and since the DDP takes care of averaging the gradient values, i do not think that this will be a problem?
> 
> I am okay with this change as it was not there to begin with. But, if this is the reason for difference between jax and pytorch, i can put this back in at a later point?

Yeah with bug I mostly meant the additional communication overhead (I only mentioned that I'm not sure if it can lead to bad behavior, because I wasn't sure if there can be issues with backpropping through an `all_reduce` operation).

But I see, so you are saying without globally calculating the loss, DDP will not match the correct global gradient. Then I'm fine with adding the syncing code back in, if you haven't seen any issues with it.

runame:
The purpose of this PR was to fix this issue, so I think it is already included?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
runame: Yeah, it doesn't hurt to round here. I'll fix it, thanks!

chandramouli-sastry:
> > The all_reduce in the loss function is not a bug, but i agree that it is a overhead. Essentially, average of normalized-ctc-losses is not equal to the globally normalized ctc-loss. On backward, all the processes will have the same gradients for all parameters and since the DDP takes care of averaging the gradient values, i do not think that this will be a problem?
> > I am okay with this change as it was not there to begin with. But, if this is the reason for difference between jax and pytorch, i can put this back in at a later point?
> 
> Yeah with bug I mostly meant the additional communication overhead (I only mentioned that I'm not sure if it can lead to bad behavior, because I wasn't sure if there can be issues with backpropping through an `all_reduce` operation).
> 
> But I see, so you are saying without globally calculating the loss, DDP will not match the correct global gradient. Then I'm fine with adding the syncing code back in, if you haven't seen any issues with it.

Yes, there is a longer discussion about that here: https://github.com/pytorch/pytorch/issues/58005

I went ahead and used dist.all_reduce, but the alternative is to use nn.all_reduce instead: but in one of the previous runs with dist.all_reduce, it seemed like there was no harm done. I am okay with this being removed as of now -- we can add this back in after carefully reviewing the effects.

znado (approved):

znado:
I will let @sourabh2k15 comment on if we should sync the CTC loss during training or not. If it is not mathematically equivalent to the single device case, we need to carefully and explicitly document this (in the function comments, in the workload module docstrings, in the README, and other places) because it is a big diff.

chandramouli-sastry:
> > > The all_reduce in the loss function is not a bug, but i agree that it is a overhead. Essentially, average of normalized-ctc-losses is not equal to the globally normalized ctc-loss. On backward, all the processes will have the same gradients for all parameters and since the DDP takes care of averaging the gradient values, i do not think that this will be a problem?
> > > I am okay with this change as it was not there to begin with. But, if this is the reason for difference between jax and pytorch, i can put this back in at a later point?
> > 
> > 
> > Yeah with bug I mostly meant the additional communication overhead (I only mentioned that I'm not sure if it can lead to bad behavior, because I wasn't sure if there can be issues with backpropping through an `all_reduce` operation).
> > But I see, so you are saying without globally calculating the loss, DDP will not match the correct global gradient. Then I'm fine with adding the syncing code back in, if you haven't seen any issues with it.
> 
> Yes, there is a longer discussion about that here: [pytorch/pytorch#58005](https://github.com/pytorch/pytorch/issues/58005)
> 
> I went ahead and used dist.all_reduce, but the alternative is to use nn.all_reduce instead: but in one of the previous runs with dist.all_reduce, it seemed like there was no harm done. I am okay with this being removed as of now -- we can add this back in after carefully reviewing the effects.

From the recent conformer-tests, we have identified that we should use dist.nn.all_reduce to match the jax implementation.

znado:
For speech (and WMT, and any workload with masked training batches), won't we still have an issue with not properly averaging the loss across devices? In the target setting (and dev) algo we take the mean of the per-device losses [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/reference_algorithms/target_setting_algorithms/jax_submission_base.py#L49), but because each of those is a weighted mean, won't this not be correct?

PR #240, opened 2022-11-03
fixing data_utils padding code

sourabh2k15:


algorithmic_efficiency/data_utils.py
```diff
@@ -23,14 +23,14 @@ def shard_and_maybe_pad_np(
   create the corresponding mask, and reshape it to be sharded across devices.
   """
   local_device_count = max(torch.cuda.device_count(), jax.local_device_count())
-  current_batch_size = batch['inputs'].shape[0]
+  current_batch_size = batch['inputs'][0].shape[0]
```
runame, line 26: Ah right, librispeech has a tuple in `batch['inputs']`. This should be something like
```python
inputs = batch['inputs']
current_batch_size = inputs[0].shape[0] if isinstance(inputs, tuple) else inputs.shape[0]
```

algorithmic_efficiency/data_utils.py
```diff
@@ -23,14 +23,14 @@ def shard_and_maybe_pad_np(
...
   remainder_size = current_batch_size % local_device_count
   if remainder_size != 0:
     if global_batch_size is not None:
       pad_size = global_batch_size - current_batch_size
     else:
       pad_size = local_device_count - remainder_size
-    targets_shape = tuple(batch['targets'].shape)
+    targets_shape = tuple(batch['targets'][0].shape)
```
runame, line 33: Same as above.

runame (approved):

algorithmic_efficiency/data_utils.py
znado: +1 we need to make sure this works for all workloads

znado (changes requested):

algorithmic_efficiency/data_utils.py
sourabh2k15: made the changes

algorithmic_efficiency/data_utils.py
sourabh2k15: done

znado (approved):

algorithmic_efficiency/data_utils.py
runame: I think I tested for all workloads, but librispeech and Criteo1TB in the original PR (didn't have the data ready).

runame: I think I tested this for all workloads, but librispeech and Criteo1TB in the original PR (didn't have the data ready).

PR #244, opened 2022-11-04
Sacrebleu and reconcile WMT diffs

priyakasimbeg:
This PR contains:
- fix to BLEU score calculation for WMT Jax, now uses sacrebleu
- fixes difference in optimizer hparams and reference hparams 
- fix in difference in MLP dim from reference implementation in Jax and Pytorch WMT workload


znado (approved):

algorithmic_efficiency/workloads/wmt/workload.py
```diff
@@ -56,7 +56,7 @@ def num_validation_examples(self) -> int:
   @property
   def num_test_examples(self) -> int:
     # wmt14_translate/de-en 'test' split size.
-    return 3003
+    return 3000
```
runame, line 59: Why are we not using the full test set anymore? Since padding has been implemented, I think this fix is not necessary anymore.

algorithmic_efficiency/workloads/wmt/workload.py
```diff
@@ -95,6 +95,8 @@ def _build_input_queue(self,
       # Without the '+1' only `num_eval_train_examples-1` examples are used
       # since one example is filtered out in the input pipeline.
       split = f'train[:{self.num_eval_train_examples+1}]'
+    if split == 'test':
+      split = f'test[:{self.num_test_examples}]'
```
runame, line 99: See my comment above.

priyakasimbeg: Sorry, will revert this

runame: No worries, I already reverted this in #246.

PR #246, opened 2022-11-04
Fix `num_eval_train_examples` for all workloads

### LINKED ISSUE #223: Choose `num_eval_train_examples` consistently for all workloads
## Description

We should choose `num_eval_train_examples` consistently for all workloads. I lean towards simply choosing it equal to `num_validation_examples`. Alternatively, we could use a fixed fraction of the train dataset, i.e. 10%.

runame:
This PR fixes #223 (+ cosmetics, i.e. change string style in some places + add more type hints).

Moreover, I undid a change introduced by #244, see [my comment](https://github.com/mlcommons/algorithmic-efficiency/pull/244#discussion_r1013771250) there.

**Issues/Questions:**
- We could just set the `num_eval_train_examples` property in `spec.py` to return `self.num_validation_examples`?
- CIFAR10 Jax was using wrong splits, I fixed this. CIFAR10 PyTorch used random subsets instead of fixed ones, even though we might want this in general, it's done differently everywhere else right now, so I changed it to be more consistent (MNIST PyTorch also shuffles right now, whereas MNIST Jax does not -- haven't changed that).
- This also raises the general question: do we want to use random subsets when we only take a subset from a split? Currently, we just use [slices](https://www.tensorflow.org/datasets/splits?hl=en). This has a few problems: 1) if the data is ordered, we might only consider a subset of classes, 2) the order might not be consistent between TF and PyTorch datasets. However, when using random subsets, it might make sense to choose a fixed random subset for both Jax and PyTorch (independent of the run's random seed) so we can better compare results across runs.
- Librispeech: according to [TF datasets](https://www.tensorflow.org/datasets/catalog/librispeech) the total number of training examples we use is larger than what we state as the `num_train_examples` property: 263.840 != 281.241 = 104.014+148.688+28.539. The same holds for the number of test examples: 'test_clean' has 2.620 samples according to TF, we say we use 2.472. And also for the validation samples: 'dev_clean' (2.703) + 'dev_other' (2.864) = 5.567 != 5348. Is this expected, maybe due to the pre-processing?
- Librispeech + FastMRI: what is the best way to restrict the eval train split to the specific number of examples when we don't use TF datasets which we can slice? If I'm not missing something we currently just restrict the number of batches we take, but this might lead to using additional training samples for evaluation.

### DIFF
--- a/algorithmic_efficiency/workloads/cifar/cifar_jax/input_pipeline.py
+++ b/algorithmic_efficiency/workloads/cifar/cifar_jax/input_pipeline.py
@@ -166,12 +166,7 @@ def create_split(split,
                  num_batches=None,
                  aspect_ratio_range=(0.75, 4.0 / 3.0),
                  area_range=(0.08, 1.0)):
-  """Creates a split from the ImageNet dataset using TensorFlow Datasets."""
-  if split == 'eval_train':
-    split = 'train'
-  elif split == 'validation':
-    split = 'test'
-
+  """Creates a split from the CIFAR-10 dataset using TensorFlow Datasets."""
   shuffle_rng, preprocess_rng = jax.random.split(rng, 2)
 
   def preprocess_example(example):
--- a/algorithmic_efficiency/workloads/cifar/cifar_jax/input_pipeline.py
+++ b/algorithmic_efficiency/workloads/cifar/cifar_jax/input_pipeline.py
@@ -201,7 +201,7 @@ def create_split(split,
   if cache:
     ds = ds.cache()
 
-  if train:
+  if train or split == 'eval_train':
     ds = ds.repeat()
     ds = ds.shuffle(16 * global_batch_size, seed=shuffle_rng[0])
 
--- a/algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
+++ b/algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
@@ -49,10 +49,8 @@ class CifarWorkload(BaseCifarWorkload):
     ds_builder.download_and_prepare()
     train = split == 'train'
     assert self.num_train_examples + self.num_validation_examples == 50000
-    if split == 'train':
+    if split in ['train', 'eval_train']:
       split = f'train[:{self.num_train_examples}]'
-    elif split == 'eval_train':
-      split = f'train[:{self.num_eval_train_examples}]'
     elif split == 'validation':
       split = f'train[{self.num_train_examples}:]'
     ds = input_pipeline.create_input_iter(
--- a/algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
+++ b/algorithmic_efficiency/workloads/cifar/cifar_jax/workload.py
@@ -48,6 +48,13 @@ class CifarWorkload(BaseCifarWorkload):
     ds_builder = tfds.builder('cifar10:3.0.2', data_dir=data_dir)
     ds_builder.download_and_prepare()
     train = split == 'train'
+    assert self.num_train_examples + self.num_validation_examples == 50000
+    if split == 'train':
+      split = f'train[:{self.num_train_examples}]'
+    elif split == 'eval_train':
+      split = f'train[:{self.num_eval_train_examples}]'
+    elif split == 'validation':
+      split = f'train[{self.num_train_examples}:]'
     ds = input_pipeline.create_input_iter(
         split,
         ds_builder,
--- a/algorithmic_efficiency/workloads/cifar/cifar_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/cifar/cifar_pytorch/workload.py
@@ -2,6 +2,7 @@
 
 import contextlib
 import math
+import random
 from typing import Dict, Optional, Tuple
 
 import torch
@@ -67,6 +68,7 @@ class CifarWorkload(BaseCifarWorkload):
         transform=transform)
     assert self.num_train_examples + self.num_validation_examples == 50000
     indices = list(range(50000))
+    random.Random(data_rng[0]).shuffle(indices)
     indices_split = {
         'train': indices[:self.num_train_examples],
         'validation': indices[self.num_train_examples:],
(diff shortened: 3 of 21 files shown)


znado:
> Moreover, I undid a change introduced by #244, see [my comment](https://github.com/mlcommons/algorithmic-efficiency/pull/244#discussion_r1013771250) there.

Good catch, thanks.

> * We could just set the `num_eval_train_examples` property in `spec.py` to return `self.num_validation_examples`?

We could, but given that we already have the separate properties, lets just keep them separate for now (in case we wanna change it again in the future too).

> * This also raises the general question: do we want to use random subsets when we only take a subset from a split? Currently, we just use [slices](https://www.tensorflow.org/datasets/splits?hl=en). This has a few problems: 1) if the data is ordered, we might only consider a subset of classes, 2) the order might not be consistent between TF and PyTorch datasets. However, when using random subsets, it might make sense to choose a fixed random subset for both Jax and PyTorch (independent of the run's random seed) so we can better compare results across runs.

I agree, lets use random subsets, and document in the README that it will be impossible to guarantee the same split between the two frameworks unless we use the same data pipeline. Hopefully if we use a large enough random sample we will get similar performance? I do think we should derive the seed from the run random seed just so we have as much determinism as possible. 

> * Librispeech: according to [TF datasets](https://www.tensorflow.org/datasets/catalog/librispeech) the total number of training examples we use is larger than what we state as the `num_train_examples` property: 263.840 != 281.241 = 104.014+148.688+28.539. The same holds for the number of test examples: 'test_clean' has 2.620 samples according to TF, we say we use 2.472. And also for the validation samples: 'dev_clean' (2.703) + 'dev_other' (2.864) = 5.567 != 5348. Is this expected, maybe due to the pre-processing?

@sourabh2k15 to look into this

> * Librispeech + FastMRI: what is the best way to restrict the eval train split to the specific number of examples when we don't use TF datasets which we can slice? If I'm not missing something we currently just restrict the number of batches we take, but this might lead to using additional training samples for evaluation.

For this, lets just set the num eval train examples to be the next multiple of the eval batch size (even if that is slightly larger than  the validation size)?

runame:
> I agree, lets use random subsets, and document in the README that it will be impossible to guarantee the same split between the two frameworks unless we use the same data pipeline. Hopefully if we use a large enough random sample we will get similar performance? I do think we should derive the seed from the run random seed just so we have as much determinism as possible.

Ok, I can do this change in a separate PR, so we can merge this one already after you have reviewed it (seems easier to review the necessary changes separately). Do you know off the top of your head what the most convenient way is to choose random subsets from TF datasets? I'm not really familiar with them, seems a bit harder since they are not indexable like in PyTorch.

> For this, lets just set the num eval train examples to be the next multiple of the eval batch size (even if that is slightly larger than the validation size)?

Yeah, that's definitely easiest implementation wise, so I'm ok with that. Will fix this.

znado:
> Ok, I can do this change in a separate PR, so we can merge this one already after you have reviewed it (seems easier to review the necessary changes separately). Do you know off the top of your head what the most convenient way is to choose random subsets from TF datasets? I'm not really familiar with them, seems a bit harder since they are not indexable like in PyTorch.

One thing would be to take X batches after shuffling, or we can generate a random int and use it to slice the dataset like:
```python
start_idx = random(0, train_size - num_examples, seed=shuffle_rng[0])
slice = f'train[{start_idx}:{start_idx+num_examples}']
```

But this isn't truly randomly shuffled, because like you said if the data is ordered it won't fix it, so maybe we should just grab X batches and make all num eval train examples a multiple of eval batch size?

runame:
> (...) so maybe we should just grab X batches and make all num eval train examples a multiple of eval batch size?

Yeah, that's an option if we also shuffle for 'eval_train'. I guess this assumes that the shuffle buffer == dataset size.

sourabh2k15:
for librispeech split sizes, our counts are lower than tfds page because we preprocess and discard examples with audio_length > 320000 and target_length > 256

znado:
> Yeah, that's an option if we also shuffle for 'eval_train'. I guess this assumes that the shuffle buffer == dataset size.

I'm assuming that shuffle buffer is large enough that we have at least some meaningful randomness in the sample. For the datasets where we also shuffle the input files as well, I think that this should be true. But you're right, we will need to make sure shuffling is turned on for 'eval_train' too. Now that I've thought through it more, I think this may be the best route to take to ensure a random sampling, even if it is a bit more work.

runame:
> I'm assuming that shuffle buffer is large enough that we have at least some meaningful randomness in the sample. For the datasets where we also shuffle the input files as well, I think that this should be true. But you're right, we will need to make sure shuffling is turned on for 'eval_train' too. Now that I've thought through it more, I think this may be the best route to take to ensure a random sampling, even if it is a bit more work.

I agree, it's probably fine if the buffer is just large enough. Should I implement this in this PR or in a separate one?

znado:
Lets do it in this PR just to save people from merging again, if that's ok with you, if you'd much rather do it in a follow-up I don't feel strongly either way.

runame:
Ok, one more issue which doesn't matter for the eval train split, but when we split the validation set out of the train split: we cannot just split by using slicing for the reasons above, but we also can't just shuffle for both and take `num_batches`, because this doesn't guarantee that the two splits are disjoint. Shuffling once and taking batches for training and validation from the same shuffled data doesn't really fit well in our current flow I think.

znado:
I honestly don't care much about MNIST and CIFAR in terms of defining conventions, so can we just do something like:
```python
train_split = 'train[:45000]'
validation_split='train[:45000]'
...
if split in ['train', 'eval_train']:
  ds = ds.shuffle()
eval_train = ds.take(10)
```
where we slice out the validation set, then only shuffle if training or doing eval_train, and then after that take the eval train batches out?

runame:
Two questions:
1. It seems like the Criteo1TB input pipeline doesn’t shuffle anywhere?
2. The librispeech input pipeline shuffles for all datasets and then shuffles again for the training split ([here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/librispeech_conformer/input_pipeline.py#L70) and [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/librispeech_conformer/input_pipeline.py#L76)), is this intended? @sourabh2k15

znado:
I'm fixing the criteo shuffling in my criteo redo PR.

Yeah that is intentional for librispeech, we effectively shuffle the files then shuffle the stream of preprocessed examples.

runame:
Ok cool, can you then shuffle for 'train' and 'eval_train' in the criteo PR? I think I have addressed everything else now.

znado (approved):

Issue #243, opened 2022-11-03
Imagenet resnet pytorch ooms


priyakasimbeg:
Imagenet Resnet Pytorch workload returns a cuda OOM. 

## Description

Traceback:
```
I1103 20:58:58.011294 140585550669632 submission_runner.py:288] Starting training loop.                                                                                            
Traceback (most recent call last):                                                                                                                                                 
  File "submission_runner.py", line 577, in <module>                                                                                                                               
    app.run(main)                                                                                                                                                                  
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/absl/app.py", line 312, in run                                                                       
    _run_main(main, args)                                                                                                                                                          
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/absl/app.py", line 258, in _run_main                                                                 
    sys.exit(main(argv))                                                                                                                                                           
  File "submission_runner.py", line 561, in main                                                                                                                                   
    FLAGS.tokenizer_vocab_path)                                                                                                                                                    
  File "submission_runner.py", line 499, in score_submission_on_workload                                                                                                           
    tokenizer_vocab_path)                                                                                                                                                          
  File "submission_runner.py", line 318, in train_once                                                                                                                             
    rng=update_rng)                                                                                                                                                                
  File "/home/kasimbeg/algorithmic-efficiency/reference_algorithms/target_setting_algorithms/pytorch_submission_base.py", line 51, in update_params                                
    loss.backward()                                                                                                                                                                
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/torch/_tensor.py", line 396, in backward                                                             
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)                                                                                             
  File "/home/kasimbeg/algorithmic-efficiency/env/lib/python3.7/site-packages/torch/autograd/__init__.py", line 175, in backward                                                   
    allow_unreachable=True, accumulate_grad=True)  # Calls into the C++ engine to run the backward pass                                                                            
RuntimeError: CUDA out of memory. Tried to allocate 2.30 GiB (GPU 0; 15.78 GiB total capacity; 10.92 GiB already allocated; 2.28 GiB free; 11.94 GiB reserved in total by PyTorch) 
[...]

PR #248, opened 2022-11-05
Jax ResNet Maxpool Padding

pomonam:
Following up on #243; This PR reverts back to our PyTorch ResNet model and changes the Jax ResNet model to ensure we have the equivalent model. 

One question @chandramouli-sastry, when I run the model diff test, it gives me:
```
Number of values with identical shapes: 161
2.682209e-07
0.0
```
Do we expect both of these terms to be 0?

### DIFF
--- a/algorithmic_efficiency/init_utils.py
+++ b/algorithmic_efficiency/init_utils.py
@@ -12,6 +12,6 @@ def pytorch_default_init(module: nn.Module) -> None:
   # Perform lecun_normal initialization.
   fan_in, _ = nn.init._calculate_fan_in_and_fan_out(module.weight)
   std = math.sqrt(1. / fan_in) / .87962566103423978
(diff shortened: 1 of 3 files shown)


chandramouli-sastry:
> Following up on #243; This PR reverts back to our PyTorch ResNet model and changes the Jax ResNet model to ensure we have the equivalent model.
> 
> One question @chandramouli-sastry, when I run the model diff test, it gives me:
> 
> ```
> Number of values with identical shapes: 161
> 2.682209e-07
> 0.0
> ```
> 
> Do we expect both of these terms to be 0?

The printed values are elementwise max and min of the absolute difference between the pytorch and jax outputs: ~1e-7 can be regarded as zero difference because the test is running on CPU and CPU implementations of pytorch and jax may not be exactly identical giving rise to these differences? Thank you for this fix!

znado (approved):

PR #253, opened 2022-11-10
ogbg-test, wmt-test and libri ds fix

chandramouli-sastry:
This PR contains the ogbg-test, wmt-test and a fix to the librispeech pytorch data loading function.
Note: I used len(ds) instead of self.num_train_examples because the num_train_examples attribute is set equal to 263840 whereas len(ds) was 263838!


algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -114,24 +114,24 @@ def _build_input_queue(self,
...
       split = 'train-clean-100+train-clean-360+train-other-500'
+      eval_train = True
     elif split == 'validation':
       split = 'dev-clean+dev-other'
     elif split == 'test':
       split = 'test-clean'
 
     ds = LibriSpeechDataset(split=split, data_dir=data_dir)
-    if split == 'eval_train':
-      indices = list(range(self.num_train_examples))
```
znado, line 134: should we instead just set `num_train_examples` to 263838?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -114,24 +114,24 @@ def _build_input_queue(self,
                          global_batch_size: int,
                          num_batches: Optional[int] = None,
                          repeat_final_dataset: bool = False):
-    del data_rng
     del num_batches
     del repeat_final_dataset
     train = False
-
+    eval_train = False
```
znado, line 120: I don't think we need this (it doesn't actually save us any lines of code, just a single python compare)?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -190,7 +190,8 @@ def loss_fn(self,
               label_smoothing: float = 0.0) -> spec.Tensor:  # differentiable
     del mask_batch
     del label_smoothing
-    return self._loss_fn(label_batch, logits_batch)['average_loss']
+    l = self._loss_fn(label_batch, logits_batch)
+    return l['average_loss'], None
```
znado, line 194: can this instead be `return l['average_loss'], l['loss']` ?

znado (approved):

znado (changes requested):
just a few small comments

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: The reason I included this is because we can create a subset of ds only after ds is created? Otherwise, we would have to replicate line 132 inside each of the if-branches? Alternatively, we can avoid renaming the original function argument by giving a new name to the `split` inside the if-conditions?

znado: I just mean, can you delete `eval_train` and change L133 back to `if split == 'eval_train':`

chandramouli-sastry: Yes, we cannot do it because we are changing the function argument `split` inside the if condition :)

chandramouli-sastry:
I added in fixes to the BatchNorm and loss function in this PR itself. Once Runa's PR is merged, I will apply my fixes on top of that and that should also resolve the return value of the loss function?

chandramouli-sastry:
I have merged my changes with that of Runa's and is now ready for your review! I had to bring back `_loss_fn` as Runa's trick of recovering the `lengths` from the `average_loss` and `summed_loss` would no longer work if `average_loss` is the global average across all devices and `summed_loss` is just the sum of losses on an individual process.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
znado: Ahhh I totally missed that, my bad! I think it can be confusing to change the value of the variable we're conditioning on in an if/elif, so instead of having eval_train can we make a new var called tfds_split that we set in each branch?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
znado: will `len(ds)` be expensive? if not, then we can leave it as is, but otherwise we should hardcode it.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -188,9 +185,31 @@ def loss_fn(
       mask_batch = torch.logical_and(mask_batch, target_lengths)
     else:
       mask_batch = target_lengths
-    n_valid_examples = max(mask_batch.sum(), 1)
+    n_valid_examples = mask_batch.sum().to(per_example_losses)
     summed_loss = per_example_losses.sum()
-    return summed_loss / n_valid_examples, per_example_losses
+    if USE_PYTORCH_DDP:
+      dist_nn.all_reduce(summed_loss)
```
znado, line 191: can you leave a comment here saying that we _need_ to use `dist_nn` in order to avoid problematic division by 8?

znado (changes requested):
LGTM just a few minor comments!

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
chandramouli-sastry: Thank you, all fixed! 
Yes, len(ds) will not be expensive as it is implemented here: https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/libri_dataset.py#L31

znado (approved):
LGTM! Thanks for these fixes

PR #254, opened 2022-11-10
Fix bleu score calculation for PyTorch WMT

runame:
This PR switches the bleu score calculation to sacrebleu for PyTorch WMT, to match the one used in Jax WMT. The only difference is that for PyTorch DDP the statistics have to be synced via an `all_gather` operation at the right position. I'm currently running a test run which checks that his sync happens correctly.

_Update: The sync works fine now._


znado (approved):

znado:
I'm going to assume this passes the reference submission tests for at least some workloads.

PR #256, opened 2022-11-14
Check if optimizer state entry has state_dict() method

runame:
Currently, when creating a checkpoint of the optimizer state in PyTorch, we try to get the `state_dict()` for all keys in `optimizer_state`. However, there is at least one case where we just use a Python function and not a PyTorch lr schedule class for the learning rate schedule (see [here](https://github.com/runame/algorithmic-efficiency/blob/main/reference_algorithms/development_algorithms/wmt/wmt_pytorch/submission.py#L86-L87)). In this case there will be an `AttributeError` raised. I think this does not affect the target setting runs though.

Do you think the easy fix I implemented here is sufficient? Maybe we have to communicate to the user that all dict entries in the `optimizer_state` have to have `state_dict()` implemented to be correctly saved?

### DIFF
--- a/algorithmic_efficiency/checkpoint_utils.py
+++ b/algorithmic_efficiency/checkpoint_utils.py
@@ -136,7 +136,8 @@ def save_checkpoint(framework: str,
     model_params = model_params.state_dict()
     optimizer_state_dict = {}
     for key in optimizer_state.keys():
-      optimizer_state_dict[key] = optimizer_state[key].state_dict()
+      if hasattr(optimizer_state[key], 'state_dict'):
+        optimizer_state_dict[key] = optimizer_state[key].state_dict()
     opt_state = optimizer_state_dict
 
   checkpoint_state = dict(
--- a/algorithmic_efficiency/checkpoint_utils.py
+++ b/algorithmic_efficiency/checkpoint_utils.py
@@ -138,6 +138,11 @@ def save_checkpoint(framework: str,
     for key in optimizer_state.keys():
       if hasattr(optimizer_state[key], 'state_dict'):
         optimizer_state_dict[key] = optimizer_state[key].state_dict()
+      else:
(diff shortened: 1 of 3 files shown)


pomonam (approved):
Thank you for raising this issue! For the sample function (for WMT), I assume that the following patch fixes the problem. However, if users define their own class with an internal state (e.g., step), it will fail when loading from the checkpoint. 

I believe that as long as we document it (either to return a function that takes in the `global_step` or to extend the PyTorch scheduler), it should be fine.

runame:
Yes exactly, if users just use a Python function which calculates the lr based on `gloabl_step`, it would work correctly like it is now, but if they use a class with state it won't. I will check the README and add a comment.

znado:
I'm fine with specifying our assumptions in the README, and changing the spec opt state type to be more explicit.

algorithmic_efficiency/checkpoint_utils.py
```diff
@@ -136,7 +136,8 @@ def save_checkpoint(framework: str,
     model_params = model_params.state_dict()
     optimizer_state_dict = {}
     for key in optimizer_state.keys():
-      optimizer_state_dict[key] = optimizer_state[key].state_dict()
+      if hasattr(optimizer_state[key], 'state_dict'):
+        optimizer_state_dict[key] = optimizer_state[key].state_dict()
```
znado, line 140: I think we should do a logging.warning here (on the first call?) that we are ignoring `key`, so the user has a chance to know.

znado (approved):

algorithmic_efficiency/checkpoint_utils.py
runame: Good idea, I have added a warning.

znado (approved):

PR #258, opened 2022-11-16
torch dataloader for librispeech JAX workloads 

sourabh2k15:


algorithmic_efficiency/workloads/librispeech_conformer/workload.py
```diff
@@ -100,17 +101,36 @@ def _build_dataset(self,
...
-                                                global_batch_size,
-                                                num_batches)
-
-    for batch in iter(ds):
-      batch = data_utils.shard_and_maybe_pad_np(
-          batch, padding_value=1, global_batch_size=global_batch_size)
-      yield batch
+    ds = LibriSpeechDataset(split=split, data_dir=data_dir)
+    ds_iter_batch_size = global_batch_size
+    sampler = None
```
znado, line 106: nit: delete and just pass `sampler=None` on L112

algorithmic_efficiency/workloads/librispeech_conformer/workload.py
```diff
@@ -100,17 +101,36 @@ def _build_dataset(self,
...
-                                                train,
-                                                global_batch_size,
-                                                num_batches)
-
-    for batch in iter(ds):
-      batch = data_utils.shard_and_maybe_pad_np(
-          batch, padding_value=1, global_batch_size=global_batch_size)
-      yield batch
+    ds = LibriSpeechDataset(split=split, data_dir=data_dir)
+    ds_iter_batch_size = global_batch_size
```
znado, line 105: nit: delete and just pass `batch_size=global_batch_size` on L110

algorithmic_efficiency/workloads/librispeech_conformer/workload.py
```diff
@@ -100,17 +101,36 @@ def _build_dataset(self,
...
+        batch_size=ds_iter_batch_size,
+        shuffle=train,
+        sampler=sampler,
+        num_workers=4,
+        prefetch_factor=10,
+        pin_memory=False,
+        drop_last=train,
+    )
+
+    dataloader = data_utils.cycle(
```
znado, line 119: given that the kwargs default to false, can you just make this `data_utils.cycle(dataloader)` ?

znado (changes requested):
Nice! Two comments:
- can we merge the `_build_dataset` into `_build_input_queue`? it doesn't look like it's actually called anywhere
- we should delete the current `algorithmic_efficiency/workloads/librispeech_conformer/input_pipeline.py` and move the pytorch one to be there (and fix imports to it) so that we don't have an unused data loader laying around (if we need the code in the future we can always go back in git history)

algorithmic_efficiency/workloads/librispeech_conformer/workload.py
sourabh2k15: done

algorithmic_efficiency/workloads/librispeech_conformer/workload.py
sourabh2k15: done

znado (approved):

PR #260, opened 2022-11-18
Remove `reverse_translation` option

### LINKED ISSUE #236: WMT Jax does not reach the target
WMT Jax does not reach the target in the expected number of steps.
## Description
Target Bleu score: 30.879
Target number of steps: 100K

Actual Bleu score: 26.558
Actual Number of steps: 148K

## Steps to Reproduce
on kasimbeg-5  on branch kasimbeg/sacrebleu ran:
```
    --framework=jax \
    --data_dir=/home/kasimbeg/tensorflow_datasets \
    --experiment_dir=/home/kasimbeg \
    --experiment_name=target_setting \
    --workload=wmt \
    --submission_path=reference_algorithms/target_setting_algorithms/jax_adamw.py \
    --tuning_search_space=reference_algorithms/target_setting_algorithms/wmt/tuning_search_space.json > results_wmt_jax.txt 2>&1
```

runame:
Removes the option for changing the direction of translation (we don't use reverse translation for the target setting runs). Fixes #236.

The target setting run of the Jax WMT workload now reaches a validation bleu score of 30.62 after about 100k steps (the target is 30.879). This run used sacrebleu 1.3.1 (should be equivalent to 1.3.0). I will add the result of the PyTorch run here when it has finished.

I might change the version of sacrebleu in a follow-up PR, but it does not seem like the latest version yields a different result from 1.3.0 (used for the target setting runs).

### DIFF
--- a/algorithmic_efficiency/workloads/wmt/input_pipeline.py
+++ b/algorithmic_efficiency/workloads/wmt/input_pipeline.py
@@ -23,13 +23,9 @@ TFDS_SPLIT_NAME = {
 }
 
 
-def normalize_feature_names(ds_info, reverse_translation,
-                            features: Features) -> Features:
+def normalize_feature_names(ds_info, features: Features) -> Features:
   """Normalizes feature names to 'inputs' and 'targets'."""
-  in_lang, tar_lang = ds_info.supervised_keys
-  input_lang = tar_lang if reverse_translation else in_lang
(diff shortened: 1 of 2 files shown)


znado (approved):

znado:
I think we should still roll back the sacrableu version to 1.3.0 just to be more consistent.

PR #261, opened 2022-11-18
ImageNet Timing Fixes

pomonam:
This PR aims to make the general PyTorch imagenet resnet workload more efficient. Specifically, this PR:
- Removes `fast_collate` and performs normalization with `torch.transform`.
- Removes sync batch norm. 
- Removes irrelevant functions.
- Updates PyTorch version (`1.12.1` --> `1.13.0`).

I confirmed that this reduces the timing for `load data` and `update_params`. Moreover, I ran full-scale experiments and confirmed that the changes still reach the target accuracy using target-setting hyperparameters. 

One minor thing is that increasing `num_workers` in the data loader seems to have a negative effect during the evaluation. I set `num_workers=0` for evaluation data loaders and this slightly slows down the evaluation. We may need to investigate this further in the future.

### DIFF
--- a/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
@@ -125,10 +125,11 @@ class ImagenetResNetWorkload(BaseImagenetResNetWorkload):
         batch_size=ds_iter_batch_size,
         shuffle=not USE_PYTORCH_DDP and is_train,
         sampler=sampler,
-        num_workers=8,
+        num_workers=4,
         pin_memory=True,
         drop_last=is_train,
-        persistent_workers=True)
+        # persistent_workers=True
+    )
     dataloader = data_utils.PrefetchedWrapper(dataloader, DEVICE)
     dataloader = data_utils.cycle(
         dataloader,
--- a/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
@@ -122,7 +122,7 @@ class ImagenetResNetWorkload(BaseImagenetResNetWorkload):
         batch_size=ds_iter_batch_size,
         shuffle=not USE_PYTORCH_DDP and is_train,
         sampler=sampler,
-        num_workers=4,
+        num_workers=2,
         pin_memory=True,
         drop_last=is_train,
         # persistent_workers=True
--- a/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
+++ b/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
@@ -74,12 +74,8 @@ class ImagenetResNetWorkload(BaseImagenetResNetWorkload):
       return map(imagenet_v2_to_torch, itertools.cycle(np_iter))
 
     is_train = split == 'train'
-    if not is_train and use_mixup:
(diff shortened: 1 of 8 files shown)


algorithmic_efficiency/workloads/imagenet_resnet/imagenet_v2.py
```diff
@@ -37,7 +37,7 @@ def _decode_example(example: Dict[str, float]) -> Dict[str, float]:
                                                resize_size)
     return {'inputs': image, 'targets': example['label']}
 
-  ds = ds.map(_decode_example, num_parallel_calls=16)
```
znado, line 40: can we undo this change? or is it needed for some reason?

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
```diff
@@ -154,7 +155,6 @@ def init_model_fn(
     model.to(DEVICE)
     if N_GPUS > 1:
       if USE_PYTORCH_DDP:
-        model = torch.nn.SyncBatchNorm.convert_sync_batchnorm(model)
```
znado, line 157: I'm a pytorch noob so I may be misunderstanding, but from my reading of the email thread, now that we delete the sync BN from here does that mean we _have_ to use DDP in order to get BN stats to sync at all? should we throw an error if `USE_PYTORCH_DDP == False` in the imagenet workloads?

znado (approved):

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_v2.py
pomonam: Sorry, I encountered a weird problem with `num_workers` for PyTorch and this was one of my attempts at debugging. I will undo this line.

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py
pomonam: In my understanding, `DataParallel ` updates the BN stats on the default device and we don't need to use synchronization (https://discuss.pytorch.org/t/understanding-behavior-of-torch-nn-dataparallel/91676).

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_v2.py
pomonam: Fixed!

runame:
@pomonam I have noticed that the `DistributedEvalSampler` was removed. What is the reason for that? I have added it because when `drop_last=False` and `len(dataset) % N_GPUS != 0` `DistributedSampler` will add in additional examples, which leads to incorrect validation metrics (see [this discussion](https://github.com/pytorch/pytorch/issues/25162)).

pomonam:
> @pomonam I have noticed that the `DistributedEvalSampler` was removed. What is the reason for that? I have added it because when `drop_last=False` and `len(dataset) % N_GPUS != 0` `DistributedSampler` will add in additional examples, which leads to incorrect validation metrics (see [this discussion](https://github.com/pytorch/pytorch/issues/25162)).

I thought that we added the condition to raise Exceptions in those cases, but it turns out that I was wrong and we were checking the other condition. I will add this function back in. Thank you for pointing it out!

runame:
> I thought that we added the condition to raise Exceptions in those cases, but it turns out that I was wrong and we were checking the other condition. I will add this function back in. Thank you for pointing it out!

Ah I see, I have implemented padding for the workloads with tf input pipelines, so we don't have to check for this condition.

pomonam:
Fixed! I incorporated all feedback.

PR #266, opened 2022-11-22
Some but not all Criteo fixes

znado:
WIP Criteo fixes. Criteo PyTorch still does not hit the target, but these are some known issues that need to be resolved.


pomonam:
Thanks for debugging the code! I will also play around with it this week.

runame:
I have briefly looked at the code and didn't spot anything yet which could cause the OOM error. If the bug isn't fixed until after NeurIPS I can also help with debugging. Two other minor comments:
1. For the `eval_train` split we currently don't shuffle, which should be changed (see [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/criteo1tb/input_pipeline.py#L112) and [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/criteo1tb/input_pipeline.py#L120-L121)).
2. [This comment](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/criteo1tb/input_pipeline.py#L132-L133) is not up-to-date, since we do not use `itertools.cycle` anywhere in this workload.

znado:
Ignore the pylint changes, those will be fixed once https://github.com/mlcommons/algorithmic-efficiency/pull/310 goes in

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/workload.py
```diff
@@ -20,6 +20,10 @@
 
 class Criteo1TbDlrmSmallWorkload(BaseCriteo1TbDlrmSmallWorkload):
 
+  @property
+  def eval_batch_size(self) -> int:
+    return 65536
```
runame, line 25: Maybe I'm out of the loop here, but is this 1/8th of the `eval_batch_size` for the Jax version of the workload on purpose?

algorithmic_efficiency/workloads/criteo1tb/workload.py
```diff
@@ -1,3 +1,5 @@
+"""Criteo1TB DLRM workload base class."""
+import json
```
runame, line 2: I don't think this is used anywhere in this file.

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_jax/models.py
```diff
@@ -104,5 +103,9 @@ def scaled_init(key, shape, dtype=jnp.float_):
               top_mlp_input)
       if layer_idx < (num_layers_top - 1):
         top_mlp_input = nn.relu(top_mlp_input)
+      if self.dropout_rate > 0.0 and layer_idx == num_layers_top - 2:
```
runame, line 106: `self.dropout_rate` can be `None`, so in that case this line will lead to an error. I think this is why pytest currently fails, since `dropout_rate` is set to `None` by default [here](https://github.com/mlcommons/algorithmic-efficiency/pull/266/files#diff-724cda3e2dc6d8233317b4fcaa9fb047fe0804003e330ecfec013c8a43d18298R63).

runame (approved):
Besides my comments, LGTM!

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/workload.py
znado: yeah, it OOMs if I increase it above this. I assume it has to do with the way we communicate tensors in pytorch, am open to fixes if others can debug it!

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_jax/models.py
znado: good catch, fixed.

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/workload.py
runame: At what point does it run OOM exactly and does it happen during the first iteration?

PR #267, opened 2022-11-22
Fix PyTorch ImageNet preprocessing with randaugment

runame:
Some functions in `randaugment.py` don't expect `torch.float32` tensors, so we have to move the `transforms.ToTensor()` transformation behind `randaugment.RandAugment()`. Does this make sense?


pomonam (approved):
LGTM! Thank you for the catch.

PR #268, opened 2022-12-07
Conformer JAX specaug fix

sourabh2k15:
with this fix conformer jax run hits target in 88.8k steps compared to pytorch run that hits target in 87k steps 
without this fix conformer jax run takes 100k steps consistently to hit target 

run link with fix : https://tensorboard.dev/experiment/oxLx9xcRSB2t5L1Kt2CaKQ/

### DIFF
--- a/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/spectrum_augmenter.py
+++ b/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/spectrum_augmenter.py
@@ -8,7 +8,7 @@ github.com/tensorflow/lingvo/blob/master/lingvo/jax/layers/spectrum_augmenter.py
 import flax.linen as nn
 import jax
 import jax.numpy as jnp
-
+import numpy as np
 
 class SpecAug(nn.Module):
   """Layer performs masking prodecure along time and frequency axis.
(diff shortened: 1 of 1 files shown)


znado (approved):

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/spectrum_augmenter.py
```diff
@@ -17,13 +17,13 @@ class SpecAug(nn.Module):
...
-  time_mask_max_frames: int = 50
-  time_mask_max_ratio: float = 1.0
+  freq_mask_count: int = 2
+  freq_mask_max_bins: int = 27
+  time_mask_count: int = 10
+  time_mask_max_frames: int = 40
+  time_mask_max_ratio: float = 0.05
   time_masks_per_frame: float = 0.0
-  use_dynamic_time_mask_max_frames: bool = False
+  use_dynamic_time_mask_max_frames: bool = True
```
znado, line 26: These are now different from the internal values, is this intentional?

znado (changes requested):

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/spectrum_augmenter.py
sourabh2k15: internally we have default values and actual values we pass in while calling specaug layer from conformer / deepspeech models , I've just removed defaults and used the runtime values here to avoid confusion

znado:
discussed offline, changes LGTM.

znado (approved):

PR #190, opened 2022-10-14
Fix OOM issue on GCP

runame:
I went with changing the config instead of writing a bash script for now, as this requires some more thought (still think it makes sense to do later on). Fixes #112 (only affected runnning the workload on GCP, not on my setup).

Also, I added a check `if global_batch_size % N_GPUS != 0`, addressing the first point in #119, and did some clean up of line-too-long exceptions (we might want to choose a better solution, see @pomonam's comment in #119).

**Important:** I didn't have time to mention it in our last meeting, but shifting all Jax operations to CPU will impact performance of the evaluation for the WMT PyTorch workload, as we used to do the beam search with Jax on GPU. I'm currently running a timing run to see how big the impact is. I think the only other workload affected by this is FastMRI, where we use Jax to compute an eval metric. Still wanted create this PR already now, to unblock running things.
Potential fix: Ideally, find out why we can't use Jax with GPU backend on GCP. Maybe more realistically, port the Jax code to PyTorch to execute everything on GPU.
_Update: WMT evaluation with Jax GPU 4909s, with Jax CPU 6140s (timed with 4xV100 32GB and bs 128)._

**Note:** The WMT workload currently fails during evalution on the test set, because the remainder batch size % N_GPUS != 0.

PR #270, opened 2022-12-09
Speed-up beam search code for WMT PyTorch

runame:
Evaluation was significantly slower for the PyTorch version of the WMT workload compared to Jax. The main cause was that we executed the beam search decoding code on CPU with Jax, since we do not allow Jax access to GPUs when running a PyTorch workload (this issue was already known when we introduced it in PR #190). I simply ported the beam search code to PyTorch and the average evaluation speed seems to be very similar now (averaged over only three runs). The only disadvantage is that we have to be careful to change both version of the code when we want to make any changes.

I'm starting a timing run until the target is reached on GCP. We can wait with merging this until the target is reached, so we know that this doesn't introduce any subtle bug which influences the bleu score.

**Update:** The run _almost_ reached the target, it got a validation bleu score of 30.644405 after about 95k steps and the target is 30.6446. Afterwards it went slightly down again. The Jax run I have done previously also only hits the target during one evaluation and then goes slightly down again. So maybe the target is set a bit too ambitiously? Or we still have some minor discrepency to the target setting run.
Also, the PyTorch update step/data selection is slower than in Jax (the data selection for obvious reasons). Here are the profiling results after three evaluations on GCP:

**Jax:**
```
----------------------------------------------------------------------------------------------------------------------------------------------------------
|  Action                	|  Mean Duration (s)	|  Std Duration (s)	|  Num Calls      	|  Total Time (s) 	|  Percentage %   	|
----------------------------------------------------------------------------------------------------------------------------------------------------------
|  Total                 	|  -----          	|  -----          	|  9295           	|  3917.3         	|  100 %          	|
----------------------------------------------------------------------------------------------------------------------------------------------------------
|  Train                 	|  3912.7         	|  0.0            	|  1              	|  3912.7         	|  99.883         	|
|  Evaluation            	|  725.81         	|  209.02         	|  3              	|  2177.4         	|  55.585         	|
|  Update parameters     	|  0.36636        	|  0.47558        	|  4644           	|  1701.4         	|  43.433         	|
|  Initializing model    	|  10.14          	|  0.0            	|  1              	|  10.14          	|  0.25884        	|
|  Data selection        	|  0.00071533     	|  0.0074099      	|  4644           	|  3.322          	|  0.084803       	|
|  Initializing dataset  	|  1.5108         	|  0.0            	|  1              	|  1.5108         	|  0.038566       	|
|  Initializing optimizer	|  0.97191        	|  0.0            	|  1              	|  0.97191        	|  0.024811       	|
----------------------------------------------------------------------------------------------------------------------------------------------------------
```

**PyTorch:**
```
----------------------------------------------------------------------------------------------------------------------------------------------------------
|  Action                	|  Mean Duration (s)	|  Std Duration (s)	|  Num Calls      	|  Total Time (s) 	|  Percentage %   	|
----------------------------------------------------------------------------------------------------------------------------------------------------------
|  Total                 	|  -----          	|  -----          	|  7629           	|  3721.6         	|  100 %          	|
----------------------------------------------------------------------------------------------------------------------------------------------------------
|  Train                 	|  3718.5         	|  0.0            	|  1              	|  3718.5         	|  99.917         	|
|  Evaluation            	|  675.02         	|  177.34         	|  3              	|  2025.1         	|  54.413         	|
|  Update parameters     	|  0.43263        	|  0.016524       	|  3811           	|  1648.7         	|  44.301         	|
|  Data selection        	|  0.0021847      	|  0.032014       	|  3811           	|  8.3258         	|  0.22371        	|
|  Initializing model    	|  3.1876         	|  0.0            	|  1              	|  3.1876         	|  0.08565        	|
|  Initializing optimizer	|  0.00074065     	|  0.0            	|  1              	|  0.00074065     	|  1.9901e-05     	|
|  Initializing dataset  	|  6.828e-06      	|  0.0            	|  1              	|  6.828e-06      	|  1.8347e-07     	|
----------------------------------------------------------------------------------------------------------------------------------------------------------
```


znado:
Given that over 20 seeds internally we saw a BLEU score range of _roughly_ 30.35-30.75 I think that is a totally fine result!

runame:
Ok cool, should we adjust the target value to be a bit more forgiving or leave it as it is?

Otherwise, I think the changes can be reviewed and merged. Then the update step should probably be sped up as well, I think there might be some operations in the model which can be implemented more efficiently, e.g. by trying to do more in-place.

znado (approved):

PR #272, opened 2022-12-15
gpu tests + fixes

chandramouli-sastry:
* This PR includes code for training identical pytorch and jax models for 10 steps with vanilla sgd and testing that their grad_norms (during training) and eval_losses are close enough. 
* In doing so, we have to turn off all stochasticity and I added a flag to the workload constructor to enable/disable specaug for librispeech models. 
* It also includes a fix for a small error in the pytorch workload of librispeech_conformer. 
* The code includes testing for imagenet_vit, imagenet_resnet, librispeech_conformer and fastmri. As of now, the fastmri test fails as the loss averaging across processes in pytorch is not identical to the jax implementation.
* I am unable to run the jax models for wmt and deepspeech as they give some nccl errors. On the other hand, the pytorch model for ogbg runs into a cuda oom!


sourabh2k15:
Hi Chandramouli, can you sync to head and try running deepspeech ? I ran it this tuesday without issues , can you also post NCCL error here so we keep track ? 

for the metrics reported can you also include train loss ? thanks!

chandramouli-sastry:
* The NCCL error and the ogbg oom error goes away when i run it with `XLA_PYTHON_CLIENT_ALLOCATOR=platform`. 
* I edited the test to also tabulate the train losses.
* Librispeech-deepspeech tests:  while the tests are running fine, the test outputs does not pass: specifically, the models behave identically for first 2 steps and then diverge. I suspected that the issue was with batchnorm and i replaced the existing batchnorm code. But, that does not fix it as well. I also observed that 2 jax runs on the same data also diverge after 2 steps! I believe that this is because of gradients exploding/vanishing. I will test the initializations of the pytorch lstm network and then see how this test runs.

sourabh2k15:
Hi Chandramouli, tests are in good shape now so we can check in the tests and debug why deepspeech steps diverge

algorithmic_efficiency/workloads/librispeech_deepspeech/librispeech_pytorch/model.py
```diff
@@ -212,122 +210,50 @@ def forward(self, inputs, input_paddings):
...
+      masked_inp = inputs.masked_fill(mask == 0, 0)
+      sum_ = (masked_inp).sum(dim=(0, 1))
+      if USE_PYTORCH_DDP:
+        sum_ = dist_nn.all_reduce(sum_)
+        count = dist_nn.all_reduce(count)
+      mean = sum_ / count
+
+      sum_ = (torch.square(masked_inp - mean) * mask).sum(dim=(0, 1))
+      if USE_PYTORCH_DDP:
+        sum_ = dist_nn.all_reduce(sum_)
```
sourabh2k15, line 243: seems like with this change torch batchnorm is closer to custom jax ones , do conformer steps match without this change ? or is this change required for conformer steps to match ?

sourabh2k15 (commented):
looks good, left one last comment

algorithmic_efficiency/workloads/librispeech_deepspeech/librispeech_pytorch/model.py
chandramouli-sastry: I did this change for the deepspeech model; the earlier batchnorm implementation is also technically identical to this but made use of the in-built syncbatchnorm for synchronizing the batch stats.

tests/test_traindiffs.py
```diff
@@ -0,0 +1,91 @@
...
+
+class ReferenceSubmissionTest(absltest.TestCase):
+
+  def test_workload(self):
+    # pylint: disable=line-too-long, unnecessary-lambda-assignment
+    for workload in WORKLOADS:
+      name = f'Testing {workload}'
+      jax_logs = '/tmp/jax_log.pkl'
+      pyt_logs = '/tmp/pyt_log.pkl'
+      run(f'XLA_PYTHON_CLIENT_ALLOCATOR=platform python3 tests/modeldiffs/train_diff.py --workload={workload} --framework=jax --global_batch_size={GLOBAL_BATCH_SIZE} --log_file={jax_logs}',
```
znado, line 39: instead of having to go through the shell and back, I think it would be cleaner if we rewrote test_traindiffs.py (or reference_algorithm_tests.py if we merge them) to have a function that takes the inputs and returns the outputs we need here. assuming we want to keep the torch dpp setup for the test, really we just need a way to run the test in torchrun. maybe we can call this directly https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py ? unfortunately I don't think that returns anything though. really wish it was cleaner to call pytorch on a multi-device setup.

for now, can you add a TODO here describing this issue?

tests/test_traindiffs.py
```diff
@@ -0,0 +1,91 @@
...
+    'wmt',
+    'librispeech_conformer',
+    'librispeech_deepspeech',
+    'fastmri',
+    'ogbg',
+]
+GLOBAL_BATCH_SIZE = 16
+
+
+class ReferenceSubmissionTest(absltest.TestCase):
```
znado, line 31: rename the class

znado (changes requested):

tests/modeldiffs/fastmri/compare.py
```diff
@@ -50,6 +50,7 @@ def sort_key(k):
   return sd
 
 
+key_transform = None
```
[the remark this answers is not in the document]
chandramouli-sastry: I added this because I am importing the transform functions within the traindiff class and i do not transform all the keys always -- for example, it imports from the fastmri/compare.py when the workload being tested is fastmri.

tests/modeldiffs/train_diff.py
```diff
@@ -0,0 +1,441 @@
...
+     'a per-device batch size of 2 is used.'))
+flags.DEFINE_boolean('use_fake_input_queue', True, 'Use fake data examples.')
+flags.DEFINE_string('log_file', '/tmp/log.pkl', 'The log file')
+FLAGS = flags.FLAGS
+USE_PYTORCH_DDP, RANK, PYTORCH_DEVICE, N_GPUS = pytorch_utils.pytorch_setup()
+N_GPUS = max(N_GPUS, jax.local_device_count())
+tf.config.set_visible_devices([], 'GPU')
+
+
+def _make_fake_image_batch(batch_shape, data_shape, num_classes):
```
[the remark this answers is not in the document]
chandramouli-sastry: Yes, this is a very good idea -- i have started working on this... will run the tests on the vm to make sure its all fine and update this PR soon!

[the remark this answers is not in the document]
chandramouli-sastry: I have added the relevant changes into reference_algorithm_tests and ensured this change does not cause any test to break. Can you please review the latest set of files? Apologies for the long turnaround time -- thank you!

tests/modeldiffs/train_diff.py
```diff
@@ -0,0 +1,441 @@
...
+      pickle.dump({'scalars': self.scalars, 'eval_results': self.eval_results},
+                  f)
+
+
+def jax_init_optimizer(workload: spec.Workload,
+                       model_params: spec.ParameterContainer,
+                       model_state: spec.ModelAuxiliaryState,
+                       hyperparameters: spec.Hyperparameters,
+                       rng: spec.RandomState) -> spec.OptimizerState:
+  """Creates an AdamW optimizer and a learning rate schedule."""
```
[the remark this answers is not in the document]
chandramouli-sastry: Done!

tests/modeldiffs/train_diff.py
```diff
@@ -0,0 +1,441 @@
...
+
+  return jax_utils.replicate(optimizer_state), opt_update_fn
+
+
+def pytorch_init_optimizer(workload: spec.Workload,
+                           model_params: spec.ParameterContainer,
+                           model_state: spec.ModelAuxiliaryState,
+                           hyperparameters: spec.Hyperparameters,
+                           rng: spec.RandomState) -> spec.OptimizerState:
+  """Creates an AdamW optimizer and a learning rate schedule."""
```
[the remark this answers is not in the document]
chandramouli-sastry: Done!

tests/test_traindiffs.py
chandramouli-sastry: I added some info into the documentation-string of this test method, can you please confirm if that is okay?

tests/test_traindiffs.py
chandramouli-sastry: Done!

znado:
This looks great, thanks for refactoring it!

znado (approved):

PR #273, opened 2022-12-15
Modify Mixup implementation

pomonam:
This PR modifies the code for mixup augmentation so that the implementation matches that of internal code.


znado (approved):
great catch!

Issue #280, opened 2023-01-05
Save all checkpoints, not just latest


znado:
Having all checkpoints will be useful for a final data release.

pomonam:
Done with #296.

PR #296, opened 2023-01-25
Add an option to save intermediate checkpoints.

pomonam:
As the title suggests, we add a flag to save intermediate checkpoints (#280). The default behavior is set to `True`.


znado (approved):
LGTM! Thanks for adding this

PR #304, opened 2023-01-27
criteo data pipeline bug fix and docker corrections

sourabh2k15:
updating pipeline to look for csv files instead of tsv


docker/scripts/startup.sh
```diff
@@ -11,13 +11,17 @@ do
     esac
 done
 
+cd ..
+./google-cloud-sdk/bin/gsutil -m cp -r gs://mlcommons-data/criteo/criteo_parts/* data/
+# yes | python3 dataset_setup.py --data_dir=~/data --temp_dir=~/data --all=False --criteo
```
znado: delete?

docker/scripts/startup.sh
```diff
@@ -11,13 +11,17 @@ do
     esac
 done
 
+cd ..
+./google-cloud-sdk/bin/gsutil -m cp -r gs://mlcommons-data/criteo/criteo_parts/* data/
```
znado: we should only do this if the workload is criteo, right?

docker/startup.sh
```diff
@@ -15,4 +15,4 @@ cd algorithmic-efficiency
 git pull
 
 chmod a+x docker/scripts/startup.sh
-docker/scripts/startup.sh -f $FRAMEWORK -u $GIT_URL -d $DATASET -s $SUBMISSION_PATH -t $TUNING_SEARCH_SPACE -e $EXPERIMENT_NAME -w $WORKLOAD
+docker/scripts/startup.sh -f $FRAMEWORK -s $SUBMISSION_PATH -t $TUNING_SEARCH_SPACE -e $EXPERIMENT_NAME -w $WORKLOAD
```
znado, line 18: is it possible to avoid having this file, and just put these three commands inside the dockerfile? it's confusing to have docker/startup.sh and docker/scripts/startup.sh

docker/scripts/cloud-init.cfg
```diff
@@ -40,9 +40,9 @@ write_files:
       ExecStartPre=/usr/bin/docker-credential-gcr configure-docker --registries us-central1-docker.pkg.dev
       ExecStartPre=mount --bind /var/lib/nvidia /var/lib/nvidia
       ExecStartPre=mount -o remount,exec /var/lib/nvidia
-      ExecStartPre=/usr/bin/docker pull us-central1-docker.pkg.dev/training-algorithms-external/mlcommons-docker-repo/mlcommons:ogbg
+      ExecStartPre=/usr/bin/docker pull us-central1-docker.pkg.dev/training-algorithms-external/mlcommons-docker-repo/mlcommons:criteo2
```
znado, line 43: why criteo2 and not criteo?

znado (changes requested):

docker/scripts/startup.sh
sourabh2k15: done

docker/scripts/startup.sh
sourabh2k15: yup removed it for now, have it locally ... will add a way to pipe dataset here so we can switch

docker/startup.sh
sourabh2k15: so Dockerfile can only run scripts that it has locally which is this script 

this script clones git repo and pulls to get a remotely controlled startup script, otherwise it's too annoying to rebuild image everytime just to make it run a different command, this setup I found useful to quickly iterate and have control

docker/scripts/cloud-init.cfg
sourabh2k15: criteo was the other image I was trying (locally build dataset copy to docker image, push to artifact registry etc) that path didn't work 

will update docker tags and update this script

docker/startup.sh
sourabh2k15: once we know exactly what docker images should run on startup we can remove the git controlled startup script

sourabh2k15: spoke offline, simplified docker setup by eliminating extra startup.sh

znado (approved):

Issue #290, opened 2023-01-23
Add Heavy Ball Momentum Baseline


bilal2vec:
## Heavy Ball Momentum

- Implementation: Optax's SGD
- Tunable hyperparameters: `learning_rate, momentum`
- Fixed hyperparameters: `nesterov=False`

priyakasimbeg:
Added this in https://github.com/mlcommons/algorithmic-efficiency/pull/285 already. So we just have to copy this from reference_algorithms/target_setting to reference_algorithms/baselines

runame:
Addressed by #306.

Issue #289, opened 2023-01-23
Add Nesterov Baseline


bilal2vec:
## Nesterov

- Implementation: Optax's SGD
- Tuned hyperparameters: `learning_rate, momentum, weight_decay`
- Fixed hyperparameters: `Nesterov=True`

runame:
Addressed by #306.

Issue #288, opened 2023-01-23
Add NAdam Baseline


bilal2vec:
### NAdam Baseline

Implementation: Internal NAdam implementation
Tunable hyperparameters: `learning_rate, b1, b2, weight_decay`
Fixed hyperparameters: `epsilon=1e8, eps_root=0.0`

runame:
Addressed by #306.

Issue #287, opened 2023-01-23
Add Adam baseline


bilal2vec:
## Adam Baseline

- Implementation: Optax's AdamW
- Tunable hyperparameters: `learning_rate, b1, b2, weight_decay` 
- Fixed hyperparameters: `epsilon=1e8, eps_root=0.0`

runame:
I assume the LR schedule is the same as for the target-setting runs?

Should `label_smoothing`, `grad_clip`, and `warmup_steps` be tunable hyperparameters? And what about the same hyperparamters for the other baselines (for the ones with polynomial LR schedule also `end_factor` and `decay_steps_factor`)?

runame:
Addressed by #306.

PR #306, opened 2023-01-28
Add Heavy-ball momentum, Nesterov momentum, AdamW, and NAdamW baselines

runame:
Addresses #287, #288, #289, and #290. I chose the directory names based on the target-setting runs naming scheme.

**TODO**

- [ ] Make adjustments based on the answer to [this question](https://github.com/mlcommons/algorithmic-efficiency/issues/287#issuecomment-1407416305).
- [ ] Add hyperparameter tuning search spaces. I don't think I have access to them, maybe someone can just replace/fill in the placeholder files I have added.
- Add instructions on how to run each submission in a README, but this can only be done when the API allows for valid submissions.


baselines/adamw/pytorch/submission.py
```diff
@@ -0,0 +1,164 @@
...
+      workload.metrics_logger.append_scalar_metrics(
+          {
+              'loss': loss.item(),
+              'grad_norm': grad_norm.item(),
+          }, global_step)
+    logging.info('%d) loss = %0.3f, grad_norm = %0.3f',
+                 global_step,
+                 loss.item(),
+                 grad_norm.item())
+
```
znado, line 118: Do you think it makes sense to add `optimizer_state[‘optimizer’].zero_grad(set_to_none=True)` to the line before the `return` in all the pytorch baselines? That should more aggressively free the memory being used by grads. I'm no pytorch pro though so I'll leave it up to you

znado (approved):

baselines/adamw/pytorch/submission.py
runame: Cool idea, I haven't really seen this anywhere together with setting the gradients to 0 before the backward pass. This could in some cases maybe lead to a small slow-down because memory has to be allocated again before each update, but apparently [not always](https://discuss.pytorch.org/t/freeing-gradients-memory-after-optimizer-step/116346/4). Do we actually care about saving memory in the case of the baselines? They run with the fixed batch sizes anyway.

znado: yeah that makes sense, I was playing around with it for criteo but that was it. fine with leaving it out for now.

PR #307, opened 2023-02-02
Add LAMB baseline

### LINKED ISSUE #291: Add LAMB baseline
## LAMB 

- Implementation: Optax's LAMB
- Tunable hyperparameters: `learning_rate, b1, b2, weight_decay`
- Fixed hyperparameters: `eps=1e-8, eps_root=0.0`

pomonam:
Fixes #291 and minor docstring changes. Additionally, only compute `grad_norm` when it is actually needed.


baselines/lamb/jax/submission.py
```diff
@@ -0,0 +1,245 @@
...
+import jax.numpy as jnp
+import optax
+
+from algorithmic_efficiency import spec
+
+_GRAD_CLIP_EPS = 1e-6
+
+
+# Forked from
+# https://github.com/deepmind/optax/blob/master/optax/_src/alias.py.
```
znado, line 22: besides the default eps of 1e-8 vs optax's 1e-6, was there a reason for forking instead of calling `optax.lamb` directly below? if not, can we just call `optax.lamb(..., eps=1e-8)` ?

znado (approved):

PR #308, opened 2023-02-02
Development workloads dataset download

pomonam:
Dataset download script for development workloads. Plus, fixed data augmentations (random crop after padding) and model configuration (e.g., no max pool for earlier layer and kernel size for first conv layer) for CIFAR models.


datasets/development_dataset_setup.py
```diff
@@ -0,0 +1,79 @@
+r"""MLCommons dataset setup script for development workloads.
```
znado, line 1: would it be possible to move this into `datasets/dataset_setup.py`? I don't think it's a big deal if we download mnist and cifar along with the other datasets if we use `--all` because theyre so tiny, but you could also exclude them from that if you want.

pomonam: That makes sense -- I will merge them!

PR #311, opened 2023-02-06
Cleaning up speech datasets

znado:
- Moving librispeech dataset to the more central root librispeech workload dir because it is used by both frameworks
- Moving jax ds pipeline logic to jax class because it is overridden in pytorch anyways
- Removing imports of base class in deepspeech because they never were/no longer necessary
- Cleaning up unused flags imports
- Small README typo


sourabh2k15 (approved):
this looks good

Issue #293, opened 2023-01-23
Add Adafactor Baseline


bilal2vec:
## Adafactor

- Implementation: Internal fork of https://github.com/google/praxis/blob/main/praxis/optimizers.py, can open-source
- Fixed hyperparameters:
```
      'opt_hparams': {
          'layerwise_adaptation': False,
          'decay_method': 'adam',
          'decay_adam': 0.99,
          'decay_pow': 0.0,
          'beta1': 0.9,
          'clip_threshold': 1.0,
          'factored': True,
          'epsilon1_grad_sq_reg': 1e-30,
          'respect_skip_lp_regularization': False,
          'exclude_from_layerwise_adaptation': None,
          'per_var_learning_summary': False,
          'sort_factored_second_moment_dims': False,
          'min_dim_size_to_factor': 128,
          'multiply_by_parameter_scale': False,
          'epsilon2_param_scale_reg': 1e-3,
          'maybe_inf_to_nan': True,
      },
```
- Tuned hyperparameters: `learning_rate, beta1, decay_adam, weight_decay`

znado:
`decay_pow` can be [ignored](https://github.com/google/praxis/blob/main/praxis/optimizers.py#L2194) because we set `'decay_method': 'adam'`

znado:
Note that our internal fork should be equivalent to [`sharded_adafactor`](https://github.com/google/praxis/blob/main/praxis/optimizers.py#L2341).

runame:
Addressed by #313 and #347.

runame:
@bilal2vec `beta1` is listed as a fixed AND tuned hparam, which one is correct?

runame:
cool, thanks for the quick reply!

PR #313, opened 2023-02-09
Adafactor Implementation

chandramouli-sastry:
Addresses #293: Here is a preliminary adafactor pytorch implementation that I adapted from huggingface that i haven't run yet.  A few notes regarding the implementation:
* I assumed decay_method='adam' implies that we would be Adam in every respect excepting that we would learn the factors of the second moment. So, I did not apply bias correction to beta as proposed in the adafactor paper: is this okay?
* I have just used the last 2 dims as the dims to be factored. Should I change this to factor the biggest 2 dimensions? Is this related to the sort_factored_second_moment_dims keyword? I am also yet to incorporate the min_dim_size_to_factor value: should I consider the both dimensions or just a single dimension when comparing against this value?
* I did not understand what decay_pow refers to: is this the 'c' in the alternative $\beta(t) = 1-t^{-c}$?
* I did not use clip_threshold and epsilon2_param_scale_reg in the implementation as multiply_by_parameter_scale is False. Also, I did not understand what respect_skip_lp_regularization means?


znado:
> * I assumed decay_method='adam' implies that we would be Adam in every respect excepting that we would learn the factors of the second moment. So, I did not apply bias correction to beta as proposed in the adafactor paper: is this okay?
I think you may need to apply the bias correction like they do [here](https://github.com/google/praxis/blob/77a06808f1c35ce0d869c05f2b9247fdf28d7fe8/praxis/optimizers.py#L1802).

> * I have just used the last 2 dims as the dims to be factored. Should I change this to factor the biggest 2 dimensions? Is this related to the sort_factored_second_moment_dims keyword? I am also yet to incorporate the min_dim_size_to_factor value: should I consider the both dimensions or just a single dimension when comparing against this value?
It looks like [it's fine](https://github.com/google/praxis/blob/77a06808f1c35ce0d869c05f2b9247fdf28d7fe8/praxis/optimizers.py#L1958) to use the last 2 dims. It seems like we can [ignore](https://github.com/google/praxis/blob/main/praxis/optimizers.py#L2358) `min_dim_size_to_factor` and `sort_factored_second_moment_dims`.

> * I did not understand what decay_pow refers to: is this the 'c' in the alternative β(t)=1−t−c?
It's fine to ignore `decay_pow`.

> * I did not use clip_threshold and epsilon2_param_scale_reg in the implementation as multiply_by_parameter_scale is False.
I think we still need to implement `clip_threshold` like is done [here](https://github.com/google/praxis/blob/main/praxis/optimizers.py#L2264) (even though it's division by 1 it still could have an effect if `reduce_rms(x) > 1`), I don't think `multiply_by_parameter_scale` disables that.

`epsilon2_param_scale_reg` is fine to ignore because `multiply_by_parameter_scale=False`.

> Also, I did not understand what respect_skip_lp_regularization means?
We can ignore that.

znado:
The implementation we used for baselines is equivalent to [`sharded_adafactor`](https://github.com/google/praxis/blob/main/praxis/optimizers.py#L2341), so feel free to base the jax implementation off that!

baselines/adafactor/pytorch/submission.py
```diff
@@ -0,0 +1,312 @@
...
+              decay_adam=hyperparameters.decay_adam,
+              weight_decay=hyperparameters.weight_decay),
+  }
+  optimizer = optimizer_state['optimizer']
+  warmup = LinearLR(
+      optimizer,
+      start_factor=1e-10,
+      end_factor=1.,
+      total_iters=hyperparameters.warmup_steps)
+  target_setting_step_hint = int(0.75 * workload.step_hint)
```
znado, line 44: we shouldnt be multiplying by 0.75, you can just delete this line

baselines/adafactor/pytorch/submission.py
```diff
@@ -0,0 +1,312 @@
...
+  del model_state
+  del rng
+
+  # Create optimizer.
+  optimizer_state = {
+      'optimizer':
+          Adafactor(
+              model_params.parameters(),
+              lr=hyperparameters.learning_rate,
+              beta1=hyperparameters.beta1,
```
znado, line 34: the hparam will be called `one_minus_beta1`, and then pass it in as `beta1=hyperparameters.one_minus_beta1,`

baselines/adafactor/pytorch/submission.py
```diff
@@ -0,0 +1,312 @@
...
+              decay_adam=hyperparameters.decay_adam,
```
znado, line 35: can you rename this to `decay_adam=hyperparameters.beta2,`

znado (approved):

PR #314, opened 2023-02-09
Refactor `loss_fn` API and fix the scaling of the loss/grad in all submissions

runame:
**Changes:**
- Minor clean up of the speech dependencies
- All masking information (`np.where(targets > 0, 1, 0)`) is now folded into `batch['weights']` for the WMT workload
- Minor optimizations for WMT PyTorch input queue
- `loss_fn` now returns a dict {'summed': scalar summed loss, 'n_valid_examples': scalar number of valid examples in batch, 'per_example': 1-d array of per-example losses}
- This PR addresses @znado's comment [here](https://github.com/mlcommons/algorithmic-efficiency/pull/239#issuecomment-1314318688) and fixes the scaling of loss/grad in all submissions

@chandramouli-sastry and @sourabh2k15: Could you please specifically review the changes to the speech workloads/submissions (comments on everyhing else are of course also welcome)?


algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
```diff
@@ -297,9 +286,9 @@ def _eval_model_on_split(self,
           targets=targets.cpu().numpy(),
           target_paddings=target_paddings.cpu().numpy(),
           tokenizer=self.tokenizer)
-      l = self._loss_fn((targets, target_paddings), (logits, logits_padding))
-      summed_loss = l['loss'].sum()
-      lengths = l['lengths']
+      loss = self.loss_fn((targets, target_paddings), (logits, logits_padding))
+      summed_loss = loss['summed']
+      lengths = loss['n_valid_examples']
```
runame, line 291: Not exactly equivalent to the previously, because `n_valid_examples` will be 1 when lengths is 0. Do you think this matters?

znado (approved):
LGTM! Thanks for tackling this tricky refactor. I'd also like someone else to take a look though in case I missed anything.

algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
```diff
@@ -284,36 +276,38 @@ def _build_input_queue(self,
               'inputs_segmentation',
               'targets',
               'targets_position',
-              'targets_segmentation'
+              'targets_segmentation',
+              'weights'
```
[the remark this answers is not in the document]
runame: To be consistent and stick to the google style guide I should probably also add a trailing comma after the last element of the dict returned by `loss_fn` everywhere?

algorithmic_efficiency/workloads/wmt/workload.py
```diff
@@ -116,6 +116,11 @@ def _build_input_queue(self,
     # when _build_input_queue is called (not when next() is first called on it).
     def _input_queue_generator():
       for batch in iter(ds):
+        weights = batch.get('weights')
```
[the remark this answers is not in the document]
runame: I think using `.get()` is correct here, since we do not expect the training batches to already have the key `weights` but validation/test set batches might have it (the remainder batches are padded).

algorithmic_efficiency/workloads/ogbg/ogbg_jax/workload.py
```diff
@@ -96,9 +96,10 @@ def _binary_cross_entropy_with_mask(
     return jnp.where(mask, losses, 0.)
 
   def _eval_metric(self, labels, logits, masks):
-    loss, _ = self.loss_fn(labels, logits, masks)
+    loss = self.loss_fn(labels, logits, masks)
+    mean_loss = loss['summed'] / loss['n_valid_examples']
```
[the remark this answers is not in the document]
runame: Ok, makes sense. But I'm not sure if we may also have batch level padding during training? Since there will be no remainder there should not be padding graphs, but maybe only padding nodes and edges?

algorithmic_efficiency/workloads/wmt/wmt_pytorch/workload.py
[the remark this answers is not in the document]
runame: For consistency, I added missing trailing commas in all places I could easily find.

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
sourabh2k15: during training this will never happen cause we repeat the dataset so there's no empty examples added to reach batch size 

during eval we do add empty examples to make the batch of desired batch size

sourabh2k15: also have you run the speech e2e model tests for this PR ? just verifying it all works

chandramouli-sastry: For the librispeech tests, I just noticed that I have set the same padding lengths for all examples and hence it will not show the diff if there is any inconsistency wrt normalization across GPUs. Could you please try with the following code (to be replaced in the reference algorithm tests file -- also, I haven't run this code myself... I'll try this in the night and create a separate pr for this correction.)
```inputs = np.random.normal(size=(*batch_shape, 320000))
targets = np.random.randint(low=1, high=1024, size=(*batch_shape, 256))
inp_pad = np.arange(0,320000)[tuple([None]*len(batch_shape))]
inp_lengths = np.random.randint(low=100000,high=320000,size=(*batch_shape,1))
inp_pad = 1*(inp_pad>inp_lengths)

tgt_pad = np.arange(0,256)[tuple([None]*len(batch_shape))]
tgt_lengths = np.random.randint(low=100,high=256,size=(*batch_shape,1))
tgt_pad = 1*(tgt_pad>tgt_lengths)
fake_batch = {
          'inputs': (inputs, inp_pad),
          'targets': (targets, tgt_pad),
      }

algorithmic_efficiency/workloads/wmt/workload.py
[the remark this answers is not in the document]
znado: yep you're right

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
runame: The librispeech tests are even failing on `main` right now. I think the Jax and PyTorch training AND evaluation losses are not equivalent. PyTorch evaluation is dividing the sum the losses from all devices and batches by the accumulated number of valid examples, in the Jax evaluation the mean of the normalized losses it taken; normally the PyTorch approach would be correct, but if I understand correctly the CTC loss is averaged differently. This also holds for training. Will try to fix and update you with more details.

chandramouli-sastry: Just a small note: Librispeech Deepspeech losses are known to diverge after 2 steps. Even consecutive jax runs are not identical for deepspeech.

runame: How is this possible if the random seed is fixed?

I think the issue I have roughly described above still holds though (just from looking at the code).

chandramouli-sastry: Here's the table for the conformer run with the existing loss function without padding changes:
image

runame: Yes, I was referring to librispeech deepspeech (I don't get why it should diverge when the seed and initial weights etc are the same). Also, if you run conformer with the padding changes, I expect it to fail.

To illustrate what I mean, in PyTorch the summed per-batch loss and `lengths` are acumulated over all eval batches, at the end summed over all devices, and the final loss is global summed loss / global summed lengths, see [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py#L301-L318). In contrast, in Jax the per-batch normalized loss (summed batch loss / batch lengths) is averaged over all eval batches, see [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py#L278-L283) and [here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/librispeech_conformer/metrics.py#L21-L30).

Usually, the PyTorch way would be correct, but for the CTC loss I'm not sure, e.g. see the definition of mean reduction for the PyTorch CTC loss [here](https://github.com/pytorch/pytorch/issues/19838#issuecomment-487669445).

runame: Also, as I mentioned via email, the PyTorch conformer model is broken for our required PyTorch version on `main` (1.13.0) and only works for PyTorch 1.12.1 (at least the test, but I assume this also affect regular training).

chandramouli-sastry: For the deepspeech: I just look at the first 2 steps and ignore the rest because the problem may be attributed to gradient explosion; the inputs we feed in to the LSTM are random normal noise and the spectrogram frames would be very similar across time steps leading to this behavior. I could not find any other explanation for jax runs not being consistent with each other but I could be wrong. 

For the conformer: Yes, we need to update the MHSA class based on the new changes in 1.13. I was referring to train losses: without using dist_nn.all_reduce, we were not able to match conformer pytorch with jax (https://github.com/chandramouli-sastry/conformer-diffs3). I remember that you removed it once in the past and I added it in again :) 

For the reference test patch: Yes, we will need to multiply with 1. instead 1! In fact, it is sufficient to just introduce padding to the targets. And from what I see, there are small diffs between pytorch and jax conformers for the reference test patch in the main even though the code is exactly same as the conformer-diffs repo! 

I think we can merge this PR and then deal with the required librispeech changes in a separate PR?

runame: **For the deepspeech:** Ok, let's ignore this for now.

**For the conformer:** I agree that we can update the MHSA class in a later PR :) So you agree that the Jax and PyTorch eval losses are not quivalent (can also be seen in the images below)? Regarding the train losses, I still believe they were not equivalent before this PR as can be seen by using your reference test patch (see below).

**For the reference test patch:** Thanks for pointing out that we only have to introduce padding to the targets. I ran the patched version of the reference test for `main` and this branch. As I expected, the train loss is not the same on `main` (despite using dist_nn.all_reduce; see the first attached image), but is the same on this branch (because I adjusted how the loss/grad is calculated in Jax as well; see the second attached image). So I think this issue is fixed by this PR. I added the patch to the test in my latest commit.

results_main
results_pr

I will run all test_traindiff tests (ignoring deepspech and eval losses for the librispeech workloads), then I think we can merge this PR. TODOs for afterwards:
- Calculate eval CTC loss consistently
- Adjust conformer model for PyTorch 1.13.x

Did I miss anything?

chandramouli-sastry: Awesome, thanks!

runame: Turns out that the test also seems to fail for ogbg on `main` (and this branch since I didn't change anything about it), which seems to be an orthogonal issue to [Zack's comment](https://github.com/mlcommons/algorithmic-efficiency/pull/314#discussion_r1103287900) since Jax and PyTorch should be both consistently wrong. I suggest that we tackle this issue in a separate PR, as we will do for the librispeech stuff above, because 1) it is already wrong on `main` and orthogonal to the purpose of this PR and 2) this PR already contains a lot of changes.

Here are the results of all tests on this branch. Besides the discussed issues with librispeech and ogbg, everything looks fine I think.

all_tests1
all_tests2

@znado @sourabh2k15 Should we merge this PR for now and proceed with librispeech and ogbg fixes in separate PRs, as suggested?

algorithmic_efficiency/workloads/ogbg/ogbg_jax/workload.py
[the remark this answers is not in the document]
znado: `jraph.dynamically_batch` calls `pad_with_graphs` [here](https://github.com/deepmind/jraph/blob/master/jraph/_src/utils.py#L1005) which could add some padding during training if we go over the max number of nodes or edges, right? I may be misreading the `jraph` lib code though.

znado:
I'm going to merge this and we can fix forward for conformer and ogbg (just take a look at my ogbg comment and let me know if it's right or if we can ignore it).

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py
sourabh2k15: cool that's fine by me, we can do the fixes in a follow-up PR

algorithmic_efficiency/workloads/ogbg/ogbg_jax/workload.py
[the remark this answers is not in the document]
runame: I think you are right. So there seem to be two different issues with ogbg:
1. Eval/train losses and grads are not the same for Jax and PyTorch in the traindiffs test.
3. The evaluation loss is not averaged correctly (your comment here; for training the loss average should be correct due to this PR).

PR #316, opened 2023-02-10
Early Stopping

danielsnider:
Stop training when a monitored metric has stopped improving.

Usage example:
```
$ python submission_runner.py \
    --early_stopping_config="early_stopping_config.json"
```

Config dict example:
```
{
  "metric_name": "validation/loss",
  "min_delta": 0,
  "patience": 0,
  "min_steps": 0,
  "max_steps": null,
  "mode": "min",
  "baseline": null
}
```

Config:
- `metric_name`: str, metric to track. "validation/loss", "train/accuracy", etc.
- `min_delta`: float, minimum change in the monitored quantity to qualify as an improvement, i.e. an absolute change of less or equal to than min_delta, will count as no improvement. Defaults to 0.
- `patience`: int, number of model evaluations with no improvement after which training will be stopped. Defaults to 0.
- `min_steps`: int, stop is never requested if step count is less than this value. Defaults to 0.
- `max_steps`: int, always stop if step count is greater than this value.
- `mode`: One of {"min", "max"}. In min mode, training will stop when the quantity monitored has stopped decreasing. In "max" mode it will stop when the quantity monitored has stopped increasing.
- `baseline`: float, baseline value for the monitored quantity. Training will stop if the model doesn't show improvement over the baseline.

This commit is based it off examples from [Jax](https://github.com/google-research/google-research/blob/ccc94ce348360ddcd41c749d4088d468ccfd1eaf/ime/utils/tools.py), [Tensorflow](https://github.com/google-research/google-research/blob/ccc94ce348360ddcd41c749d4088d468ccfd1eaf/simpdom/run_node_level_model.py), and [Keras](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping).

### DIFF
--- a/algorithmic_efficiency/early_stopping.py
+++ b/algorithmic_efficiency/early_stopping.py
@@ -1,0 +1,106 @@
+import json
+
+from absl import logging
+import numpy as np
+
+
+class EarlyStopping:
+  """Stop training early if not improving.
+
+  Args:
+    config: A dict containing the arguments for this class or a str of a path
+      to a JSON file containing the config.
+
+  Usage example:
+  ```
+  $ python submission_runner.py \
+      --early_stopping_config="early_stopping_config.json"
+  ```
+
+  Config dict example:
+  ```
+  {
+    "metric_name": "validation/loss",
+    "min_delta": 0,
+    "patience": 0,
+    "min_steps": 0,
+    "max_steps": null,
+    "mode": "min",
+    "baseline": null
+  }
+  ```
+
+  Config:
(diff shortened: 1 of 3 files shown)


znado:
Closing this as we discussed via email.

Note: I don't think we want to put this in the main submission_runner.py, in case it interferes with something a submitter is doing. **But** I think we could have a self-tuning submission that does something like this, or may more generally a self-tuning submission that (intelligently) tries different hparams and stops them early if they aren't doing well (sort of like a bayesian optimization loop with a single worker).

PR #320, opened 2023-02-15
Development workload updates

pomonam:
Duplicate #308 for conflicts. For context, we weren't hitting the target previously, but with this change, we are able to hit the target.

I merged the data downloading scripts following @znado suggestions. 

Regarding the comment, https://github.com/mlcommons/algorithmic-efficiency/pull/308#discussion_r1099233883, I changed `train_mean` and `train_std` accordingly to reflect this change.


algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/input_pipeline.py
```diff
@@ -152,7 +152,9 @@ def _decode_and_center_crop(image_bytes: spec.Tensor,
 
 def normalize_image(image: spec.Tensor,
                     mean_rgb: Tuple[float, float, float],
-                    stddev_rgb: Tuple[float, float, float]) -> spec.Tensor:
+                    stddev_rgb: Tuple[float, float, float],
+                    dtype=tf.float32) -> spec.Tensor:
+  image = tf.image.convert_image_dtype(image, dtype)
```
znado, line 157: why do we need this? also, was there supposed to be a change to the [train mean/stddev](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/imagenet_resnet/workload.py#L59) included in this PR?

pomonam: I added this line as I was receiving an error message that `dtype` is incorrect. I can further look into this issue and see if this could be avoided. 

Yes, I changed `train_mean/std` in this PR (divided the original value by 255).

znado: I may totally be misreading the GH PR review UI, so let me know if I am just missing the link to it, but I don't see the changes to the imagenet train mean/stddev included in this PR, only the CIFAR ones https://github.com/pomonam/algorithmic-efficiency/blob/development_workload/algorithmic_efficiency/workloads/imagenet_resnet/workload.py#L59

at least for imagenet, we already call this right after this function returns [here](https://github.com/mlcommons/algorithmic-efficiency/blob/7d0d628fdfb867a3bba0621c6b9d3e8bc8c839cd/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/input_pipeline.py#L205), so I think we should just do that for CIFAR and not touch this file

znado: also for TF pipelines in the future I think if you want to just change the dtype, it will be much safer to just cast to float32, because `convert_image_dtype` also rescales the values of the images (we have had multiple bugs where we accidentally rescale 0 or 2 times because of it) https://www.tensorflow.org/api_docs/python/tf/image/convert_image_dtype

znado (changes requested):

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/input_pipeline.py
pomonam: Thank you for pointing this out and for all your suggestions. I had a misunderstanding that your comment was on the mean/stddev of CIFAR. I separated the code for CIFAR and ImageNet, and this PR should no longer affect ImageNet results.

znado (approved):

Issue #317, opened 2023-02-15
PyTorch conformer model broken for PyTorch 1.13.x


runame:
The PyTorch conformer model throws the following error ([here](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/models.py#L390-L402)) when PyTorch 1.13.0 is used (which is the version in our requirements):
```
return torch._native_multi_head_attention(
RuntimeError: Mask Type should be defined
```

sourabh2k15:
this is now fixed by https://github.com/mlcommons/algorithmic-efficiency/pull/321

PR #321, opened 2023-02-16
Conformer update

chandramouli-sastry:
Updates the Conformer MHSA to Pytorch 1.13.1: addresses https://github.com/mlcommons/algorithmic-efficiency/issues/317 and https://github.com/mlcommons/algorithmic-efficiency/issues/340

### DIFF
--- a/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/models.py
+++ b/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/models.py
@@ -291,8 +291,7 @@ class MHSAwithQS(nn.MultiheadAttention):
             to ignore for the purpose of attention (i.e. treat as "padding"). For unbatched `query`, shape should be :math:`(S)`.
             Binary and byte masks are supported.
             For a binary mask, a ``True`` value indicates that the corresponding ``key`` value will be ignored for
-            the purpose of attention. For a byte mask, a non-zero value indicates that the corresponding ``key``
-            value will be ignored.
+            the purpose of attention. For a float mask, it will be directly added to the corresponding ``key`` value.
         need_weights: If specified, returns ``attn_output_weights`` in addition to ``attn_outputs``.
             Default: ``True``.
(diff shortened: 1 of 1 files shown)


algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/models.py
```diff
@@ -399,7 +408,9 @@ def forward(self,
             self.out_proj.bias,
             key_padding_mask if key_padding_mask is not None else attn_mask,
             need_weights,
-            average_attn_weights)
+            average_attn_weights,
+            1 if key_padding_mask is not None else
+            0 if attn_mask is not None else None)
```
znado, line 413: nit: can you pull this out into a named variable? the nested ternary statements are a bit hard to read :)

znado (approved):

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/models.py
chandramouli-sastry: Yes, that's true! But, I left them in because thats how they are written in the pytorch 1.13 codebase and it will be easier to compare and incorporate diffs in the future?

pomonam:
Nit: We might want to use single quote for consistency?

chandramouli-sastry:
> Nit: We might want to use single quote for consistency?

Hi! Yeah, but pytorch uses double quotes and i left them as is so that it makes it easier to compare and update if the attention module gets updated again in pytorch? I don't know what is the right thing to do in this particular case -- any thoughts?

runame:
What part of `nn.MultiheadAttention` is modified here? Maybe there is a way to avoid copying the source code completely...

chandramouli-sastry:
> What part of `nn.MultiheadAttention` is modified here? Maybe there is a way to avoid copying the source code completely...

The bias corresponding to query vector and the query vector are scaled by the QueryScaler module.

Issue #331, opened 2023-03-02
Imagenet ViT broken w Incompatible shapes for broadcasting


priyakasimbeg:
Imagenet ViT broken w Incompatible shapes for broadcasting

## Description

Traceback
```

Traceback (most recent call last):
  File "submission_runner.py", line 599, in <module>
    app.run(main)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "submission_runner.py", line 572, in main
    score = score_submission_on_workload(workload,
  File "submission_runner.py", line 507, in score_submission_on_workload
    timing, metrics = train_once(workload, global_batch_size,
  File "submission_runner.py", line 320, in train_once
    optimizer_state, model_params, model_state = update_params(
  File "/algorithmic-efficiency/reference_algorithms/target_setting_algorithms/jax_submission_base.py", line 98, in update_params
    new_optimizer_state, new_params, new_model_state, loss, grad_norm = pmapped_train_step( # pylint: disable=line-too-long
  File "/algorithmic-efficiency/reference_algorithms/target_setting_algorithms/jax_submission_base.py", line 50, in pmapped_train_step
    (summed_loss, (n_valid_examples, new_model_state)), grad = grad_fn(
  File "/algorithmic-efficiency/reference_algorithms/target_setting_algorithms/jax_submission_base.py", line 40, in _loss_fn
    loss_dict = workload.loss_fn(
  File "/algorithmic-efficiency/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/workload.py", line 179, in loss_fn
    per_example_losses *= mask_batch
  File "/usr/local/lib/python3.8/dist-packages/jax/_src/numpy/lax_numpy.py", line 4639, in deferring_binary_op
    return binary_op(*args)
  File "/usr/local/lib/python3.8/dist-packages/jax/_src/numpy/ufuncs.py", line 80, in fn
    x1, x2 = _promote_args(numpy_fn.__name__, x1, x2)
  File "/usr/local/lib/python3.8/dist-packages/jax/_src/numpy/util.py", line 344, in _promote_args
    return _promote_shapes(fun_name, *_promote_dtypes(*args))
[...]

PR #346, opened 2023-03-10
Speech workloads mask fix

sourabh2k15:
in speech workload cases mask is already given by target_paddings , it's a 2d array containing paddings for individual examples as well as the batch dimension

### DIFF
--- a/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
+++ b/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
@@ -76,7 +76,7 @@ class LibriSpeechConformerWorkload(workload.BaseLibrispeechWorkload):
       }
 
       padded_batch = data_utils.shard_and_maybe_pad_np(
-          numpy_batch, padding_value=1.0, global_batch_size=global_batch_size)
+          numpy_batch, padding_value=1, global_batch_size=global_batch_size)
       print(padded_batch['weights'])
       yield padded_batch
 
--- a/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
+++ b/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
@@ -169,8 +169,8 @@ class LibriSpeechConformerWorkload(workload.BaseLibrispeechWorkload):
                                        target_paddings)
     # mask_batch is assumed to be shape [batch].
     if mask_batch is not None:
-      per_example_losses *= mask_batch
       mask_batch = jnp.logical_and(mask_batch, 1 - target_paddings)
+      per_example_losses *= mask_batch
     else:
       mask_batch = 1 - target_paddings
     n_valid_examples = jnp.maximum(mask_batch.sum(), 1)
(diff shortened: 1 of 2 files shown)


algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
```diff
@@ -166,10 +166,11 @@ def loss_fn(
                                        logit_paddings,
                                        targets,
                                        target_paddings)
-    # mask_batch is assumed to be shape [batch].
+    # mask_batch is assumed to be shape [batch] for other workloads
+    # but in speech cases it's [batch, target_len].
     if mask_batch is not None:
-      per_example_losses *= mask_batch
       mask_batch = jnp.logical_and(mask_batch, 1 - target_paddings)
```
runame, line 172: I think we multiplied by `mask_batch` after this line because `target_paddings` are already included in the ctc loss calculation, but I think it should not make a difference?

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
```diff
@@ -166,10 +166,11 @@ def loss_fn(
...
+      per_example_losses *= mask_batch[:, 0]
```
runame, line 173: Do we expect the mask to be the same across the second dimension (of size target_len) or why is it correct to only use the first slice of the second dimension?
If the size of `mask_batch` is (batch_size, target_len), can't we just also pass it together with target_paddings to `self.ctc_loss`? If we just want to provide the user the option to mask out the contribution of specific examples to the loss, the shape of `mask_batch` should maybe just be (batch_size,).

algorithmic_efficiency/workloads/librispeech_conformer/librispeech_jax/workload.py
```diff
@@ -166,10 +166,11 @@ def loss_fn(
...
     else:
       mask_batch = 1 - target_paddings
     n_valid_examples = jnp.maximum(mask_batch.sum(), 1)
```
runame, line 176: If the shape of `mask_batch` is (batch_size, target_len), do we really want to sum over both dimensions to get `n_valid_examples` or should it just be the number of examples for which mask batch is not all zero?

runame (commented):
Some more questions/comments:
1. This logic did not change in any recent refactor AFAIK, so I'm confused why it only know starts to make problems. Maybe we just didn't run the speech workloads since we introduced this logic? I guess the tests didn't fail because they do not provide an extra mask?
2. Is the PyTorch version of the workload not also affected?
3. ImageNet ViT is also affected by this (see #331) and it worked fine before. I suspect the issue is caused by something non-specific to any individual workload and should be fixed on that level.

runame:
I think I have found a potential cause of the bug, [this commit](https://github.com/mlcommons/algorithmic-efficiency/commit/8d3fd81d9e5fcecb0435fee6faa9ff591f5460d3) creates weights by default and not only when they are needed due to padding being applied. [This line](https://github.com/mlcommons/algorithmic-efficiency/blob/main/algorithmic_efficiency/data_utils.py#L41) will create a 2d mask even for ImageNet for which this is not desired.

@pomonam I guess you wanted to be able to always broadcast the weights when using a shared input pipeline? I guess we can still achieve this with a few changes.

pomonam:
@runame Thank you for debugging this! At the time of the commit, I don't think `mask_shape = targets_shape if len(targets_shape) < 3 else targets_shape[0]` was triggered, since I remember running imagenet workloads. But, as you suggested, an easy fix here for now would be to revert these lines (indent them again).

Yes, I wanted to broadcast the weights when using the shared input pipeline (also for imagenet-v2) and I also agree that we only need mask during eval. I can make changes on other places to support this, so how about we go ahead revert my commits and then I will change workloads (e.g., mnist) that relied on my commit?

pomonam:
#349 should consume this PR.

Issue #358, opened 2023-03-23
Make `goal_reached` condition consistent with rules


runame:
## Description

Currently, we only stop training before the time/step limit is reached if the [validation and test target are reached in the latest evaluation](https://github.com/mlcommons/algorithmic-efficiency/blob/main/submission_runner.py#L368-L370). However, according to the rules the condition should be met as soon as both targets have been hit _at some point_, i.e. hitting the targets in different evaluations should also flip the `goal_reached` bit.

runame:
Closed by #360.

PR #360, opened 2023-03-25
Fix for if a submission has reached both targets

znado:
Fixes https://github.com/mlcommons/algorithmic-efficiency/issues/358


pomonam (approved):
LGTM!

Issue #292, opened 2023-01-23
Add Shampoo baseline


bilal2vec:
## Shampoo

- Implementation: https://github.com/google-research/google-research/blob/master/scalable_shampoo/optax/distributed_shampoo.py
- Fixed/default hyperparameters:
```
      'opt_hparams': {
          'block_size': 1024,
          'beta1': 0.9,
          'beta2': 0.999,
          'diagonal_epsilon': 1e-8,
          'matrix_epsilon': 1e-6,
          'start_preconditioning_step': 101,
          'preconditioning_compute_steps': 20,
          'statistics_compute_steps': 1,
          'graft_type': 4,
          'best_effort_shape_interpretation': True,
          'best_effort_memory_usage_reduction': True,
          'nesterov': True,
          'exponent_override': 0,
          'mesh_axis_names': None,
          'num_devices_for_pjit': None,
          'shard_optimizer_states': False,
          'inverse_failure_threshold': 0.1,
          'moving_average_for_momentum': True,
          'skip_preconditioning_dim_size_gt': 2048,
          'clip_by_scaled_gradient_norm': None,
          'eigh': True,
      },
```
Tunable hyperparameters: `learning_rate, beta1, beta2, weight_decay`

runame:
Just to mention it here, we don't have a PyTorch implementation yet.

PR #361, opened 2023-03-29
Add Shampoo Jax baseline submission

runame:
Addresses part of #292. I couldn't find the `mesh_axis_names` hparam mentioned in the issue.


znado (approved):
Approving so we can merge, will fix forward any issues

PR #366, opened 2023-03-30
ogbg and conformer fixes

chandramouli-sastry:
This PR includes the following:
* I fixed the query scaling in the conformer to scale both projection weight and bias -- earlier, I was scaling the bias and the query before the projection weight was applied. I tested this by using non-trivial initial values to make sure that the implementation is correct. 
* For the OGBG workloads, I changed the eval methods and pass the per-example-losses instead of average -- metrics.Average can now take care of masking.
* The traindiff for OGBG workloads was arising from different epsilon values in the layernorm layers -- after correcting the epsilon to 1e-6 from 1e-5, the difference between jax and pytorch runs is negligible! As of now, deepspeech is the only workload for which the traindiffs diverge after 1 gradient step.


algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/models.py
```diff
@@ -264,6 +264,24 @@ def __init__(self, config: ConformerConfig):
         batch_first=True)
     self.qs = QueryScaler(dim=config.encoder_dim // config.num_attention_heads)
 
+  def _scaled_in_proj_weight(self):
+    # Scale the query projection weight.
+    in_proj_queryW_scaled = self.qs(self.in_proj_weight[:self.embed_dim].view(
```
runame, line 269: Nit: for better readability we could split this in two different expressions, e.g.
```python
qs_input = self.in_proj_weight[:self.embed_dim].view(
    self.num_heads, self.embed_dim // self.num_heads, -1).transpose(1, 2)
# Scale the query projection weight.
in_proj_queryW_scaled = self.qs(qs_input).transpose(1, 2).view(
    *self.in_proj_weight[:self.embed_dim].shape)
```

runame (approved):
LGTM! Thanks a lot for fixing these things.

znado (approved):

PR #369, opened 2023-04-13
Add hparams logging with wandb

runame:
Unless I'm missing something `wandb` is currently not logging the hyperparameters. I have added this functionality; I'm just starting to use `wandb`, so please let me know if there is a better way to do this.


pomonam (approved):
LGTM! This would be useful for tuning hyperparams.

PR #390, opened 2023-05-12
Various version updates & have an option to apply `torch.compile`

pomonam:


pomonam:
Making this as draft for now, because we need to confirm backward compatability across all workloads.

pomonam:
I will update the code in #405

PR #396, opened 2023-05-30
timing changes

priyakasimbeg:
Add torch.cuda.synchronize() each time we use time.time().
Remove train_start_time measurement.

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -168,6 +168,18 @@ flags.DEFINE_boolean('save_checkpoints',
 FLAGS = flags.FLAGS
 USE_PYTORCH_DDP, RANK, DEVICE, N_GPUS = pytorch_setup()
 
+def _get_time():
+  return time.time()
+
+def _get_time_ddp():
+  torch.cuda.synchronize()
+  t = time.time()
+  return sync_ddp_time(t, DEVICE)
+
+if USE_PYTORCH_DDP:
+  get_time = _get_time_ddp
+else:
+  get_time = _get_time
 
 def convert_filepath_to_module(path: str):
   base, extension = os.path.splitext(path)
@@ -315,7 +327,7 @@ def train_once(
 
   if USE_PYTORCH_DDP:
     torch.cuda.synchronize()
-  global_start_time = time.time()
+  global_start_time = get_time()
   if USE_PYTORCH_DDP:
     # Make sure all processes start training at the same time.
     global_start_time = sync_ddp_time(global_start_time, DEVICE)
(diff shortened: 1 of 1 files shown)


submission_runner.py
```diff
@@ -312,10 +329,13 @@ def train_once(
                                                  hyperparameters)
     workload.attach_metrics_logger(metrics_logger)
 
-  global_start_time = time.time()
+  if USE_PYTORCH_DDP:
```
mikerabbat, line 332: Lines 332 and 333 are redundant now and can be removed, since `if USE_PYTORCH_DDP` evaluates to true, then `get_time = _get_time_ddp` which calls synchronize

mikerabbat (changes requested):
Minor changes around where `global_start_time` gets set, but otherwise looks good to me!

submission_runner.py
```diff
@@ -312,10 +329,13 @@ def train_once(
...
+    torch.cuda.synchronize()
+  global_start_time = get_time()
   if USE_PYTORCH_DDP:
```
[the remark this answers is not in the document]
priyakasimbeg: fixed

submission_runner.py
priyakasimbeg: fixed, forgot to clean that up

priyakasimbeg:
> Minor changes around where `global_start_time` gets set, but otherwise looks good to me!

fixed!

submission_runner.py
```diff
@@ -361,12 +372,10 @@ def train_once(
...
       train_state['training_complete'] = True
 
-    train_step_end_time = time.time()
-    if USE_PYTORCH_DDP:
-      train_step_end_time = sync_ddp_time(train_step_end_time, DEVICE)
+    train_step_end_time = get_time()
 
     train_state['accumulated_submission_time'] += (
-        train_step_end_time - train_step_start_time)
+        train_step_end_time - train_state['last_step_end_time'])
```
znado, line 378: does this change the semantics of `accumulated_submission_time` to include the evals? IIRC the goal of `accumulated_submission_time` was to just count the time spent in submission functions data selection and update.

znado (changes requested):

submission_runner.py
priyakasimbeg: The train_state['last_step_end_time'] is set to the time after the eval (if there is an eval) line 459. So the accumulated_submission_time should not include the eval time with the code as is now.

znado (approved):

PR #399, opened 2023-06-07
Criteo and Deepspeech Comparators

chandramouli-sastry:
This PR includes:
* An update to deepspeech comparator so that it can use the jax cudnn lstm. 
* I also updated the generation of fake test-data for deepspeech and this reduces the variance of test-results between runs and makes it closer to the test results obtained with real data samples. 
* Criteo comparator: using fake data, the test could not catch any diff. The plan is to use this comparator for running tests with a real data dump (and, possibly the appropriate optimizer for this workload instead of the default sgd).

### DIFF
--- a/algorithmic_efficiency/workloads/librispeech_deepspeech/librispeech_pytorch/models.py
+++ b/algorithmic_efficiency/workloads/librispeech_deepspeech/librispeech_pytorch/models.py
@@ -24,7 +24,7 @@ class DeepspeechConfig:
   """Global hyperparameters used to minimize obnoxious kwarg plumbing."""
   vocab_size: int = 1024
   encoder_dim: int = 512
-  num_lstm_layers: int = 4
+  num_lstm_layers: int = 6
   num_ffn_layers: int = 3
   conv_subsampling_factor: int = 2
   conv_subsampling_layers: int = 2
--- a/tests/modeldiffs/librispeech_deepspeech/compare.py
+++ b/tests/modeldiffs/librispeech_deepspeech/compare.py
@@ -32,6 +32,8 @@ def key_transform(k):
       i = i.replace('Conv1d', 'Conv')
     elif 'MHSAwithQS' in i:
       i = i.replace('MHSAwithQS', 'SelfAttention')
+    elif 'LSTM' in i:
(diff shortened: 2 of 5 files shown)


tests/test_traindiffs.py
```diff
@@ -24,6 +24,7 @@
     'librispeech_deepspeech',
     'fastmri',
     'ogbg',
+    'criteo1tb'
```
runame, line 27: Nit: add trailing comma.

runame (approved):
LGTM, thanks for implementing this!

algorithmic_efficiency/workloads/librispeech_deepspeech/librispeech_pytorch/models.py
```diff
@@ -24,7 +24,7 @@ class DeepspeechConfig:
   """Global hyperparameters used to minimize obnoxious kwarg plumbing."""
   vocab_size: int = 1024
   encoder_dim: int = 512
-  num_lstm_layers: int = 4
+  num_lstm_layers: int = 6
```
znado, line 27: nit: I think this needs to be synced because this is already 6 at HEAD

tests/modeldiffs/librispeech_deepspeech/compare.py
```diff
@@ -54,30 +56,24 @@ def sd_transform(sd):
...
+      out[k[:-1]] = l
     else:
       out[k] = sd[k]
   keys_to_del = []
+  updates = dict()
   for k in out:
-    if (k[-2] in ['ii', 'if', 'ig', 'io']) and k[-1] == 'bias':
-      other_bias = k[:-2] + ('h' + k[-2][1], 'bias')
-      out[other_bias] = out[other_bias] + out[k]
+    if isinstance(out[k], dict):
```
znado, line 67: nit: nbd if not, but is it possible to have a more descriptive check here? like checking for `k` being whatever the values of `k[:-1]` in the `elif 'LSTM' in ''.join(k):` above are? if not, can you add a comment?

znado (approved):
This is great, thanks so much for wrapping these up!

Issue #324, opened 2023-02-22
Data setup: Librispeech downloading code raises exception


priyakasimbeg:
From Mike's test:
> None of the `tar` commands succeed. This happens on [line 433-434](https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/dataset_setup.py#L433-L434) and on [line 447](https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/dataset_setup.py#L447), and I believe they fail for different reasons.
The first time ([line 433-434](https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/dataset_setup.py#L433-L434)), the path to the tar file is incorrect. I believe the script is intended to be run from the algorithmic_efficiency root folder (instructions say call `python3 datasets/dataset_setup.py …`), and I used different tmp and data directories
The second time ([line 447](https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/dataset_setup.py#L447)) it fails because wget (line 445) hasn’t completed yet
To fix this, I suggest:
passing `cwd=tmp_librispeech_dir` to the Popen constructor on line 434
appending .communicate() on line 445
change line 447 to `subprocess.Popen(f'tar xzvf {tar_filename}', shell=True, cwd=tmp_librispeech_dir).communicate()`
(I tested those changes locally and they solved those problems)
After untarring the files, everything is in tmp_librispeech_dir/LibriSpeech so [line 450](https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/dataset_setup.py#L450) should be changed to take `data_dir=os.path.join(tmp_librispeech_dir, ‘LibriSpeech’)`
There are also path-related problems in the librispeech_tokenizer.py. Here the file spm_model.vocab gets copied to the algorithmic_efficiency directory (from which `python3` was a called). Then it isn’t found when `librispeech_tokenizer.load_tokenizer()` gets called from `librispeech_preprocess.run()`

priyakasimbeg:
Open question on pr/400 from @sourabh2k15:
> I ran the script and it wasn't able to find tokenizer after it trained it , where do we write the trained tokenizer to on disk ?

chandramouli-sastry:
[...]

PR #400, opened 2023-06-08
Librispeech dataset fixes

chandramouli-sastry:
Addresses https://github.com/mlcommons/algorithmic-efficiency/issues/324 closely following @mikerabbat's suggestions. @sourabh2k15 , could you please check this and confirm that everything is in order?
The command to run is: `python3 dataset_setup.py --data_dir=~/data --librispeech`


datasets/dataset_setup.py
```diff
@@ -436,19 +436,17 @@ def download_imagenet_v2(data_dir):
 
 
 def download_librispeech(dataset_dir, tmp_dir, train_tokenizer):
+  # pylint: disable=line-too-long
```
znado, line 439: nit: can you put this just on the lines that need it? you can put it as a trailing comment at the end of the line and it will affect just that line

datasets/dataset_setup.py
```diff
@@ -436,19 +436,17 @@ def download_imagenet_v2(data_dir):
...
   # After extraction the result is a folder named Librispeech containing audio
   # files in .flac format along with transcripts containing name of audio file
   # and corresponding transcription.
-  tmp_librispeech_dir = os.path.join(tmp_dir, 'librispeech')
+  tmp_librispeech_dir = os.path.join(tmp_dir, 'LibriSpeech')
```
znado, line 443: nit: I'd prefer to keep all filepaths lowercase (to be consistent with the rest of the datasets) unless we have a good reason otherwise

datasets/dataset_setup.py
```diff
@@ -457,17 +455,14 @@ def download_librispeech(dataset_dir, tmp_dir, train_tokenizer):
       'train-other-500.tar.gz',
   ]
   for tar_filename in tars:
-    wget_cmd = (f'wget --directory-prefix={tmp_librispeech_dir} '
-                f'http://www.openslr.org/resources/12/{tar_filename}')
-    subprocess.Popen(wget_cmd, shell=True)
-    tar_path = os.path.join(tmp_librispeech_dir, tar_filename)
-    subprocess.Popen(f'tar xzvf {tar_path}', shell=True)
```
znado, line 464: just confirming, we don't need to untar the files?

datasets/librispeech_preprocess.py
```diff
@@ -167,9 +167,11 @@ def run(input_dir, output_dir, tokenizer_vocab_path):
...
   for subset in subset_list:
     logging.info('Processing split = %s...', subset)
-    subset_dir = os.path.join(output_dir, subset)
-    os.makedirs(subset_dir, exist_ok=True)
-    example_ids, num_entries = preprocess_data(subset_dir, tokenizer, subset)
+    in_dir = os.path.join(input_dir, subset)
+    out_dir = os.path.join(output_dir, subset)
+    os.makedirs(out_dir, exist_ok=True)
+    example_ids, num_entries = preprocess_data(in_dir, output_dir, tokenizer,
+    subset)
```
znado, line 174: nit: fix the arg formatting, something like:
```
example_ids, num_entries = preprocess_data(
    in_dir, output_dir, tokenizer, subset)
```

datasets/dataset_setup.py
chandramouli-sastry: The untar now happens together with the wget.

datasets/dataset_setup.py
chandramouli-sastry: As the comment above says, the extracted audio files is inside Librispeech and this is the data_dir that the tokenizer expects.

sourabh2k15:
Hi chandramouli, code looks good but where does the user run commands from ? do they cd into datasets and run `python3 dataset_setup.py --data_dir=~/data --librispeech` or do they run it from algo efficiency root dir ? 

asking as that can lead to error related to tokenizer file path

sourabh2k15:
also if train_tokenizer is False user should provide tokenizer_vocab_path to be used in preprocessing script , current code assumes train_tokenizer will always be True ?

chandramouli-sastry:
> 

Hi! Thanks so much for checking -- all paths are absolute, so there should be no issue. I just ran it once from outside the datasets directory and there was no issue. Did you notice any relative path anywhere? 

Are you referring to the hard-coded call to `download_librispeech` (https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/dataset_setup.py#L551) even though there's a train_tokenizer flag here (https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/dataset_setup.py#L166)? I think we should simply remove the train_tokenizer argument to the download_librispeech function and also remove the flag -- thoughts? 
Otherwise, we should add a flag for the user to provide a vocab_path -- i didn't know if this would be particularly useful?

datasets/librispeech_preprocess.py
chandramouli-sastry: Done!

sourabh2k15:
I'm fine with removing train_tokenizer flag if we write the trained tokenizer to spm_model.vocab as user needs to pass that in full training run

sourabh2k15:
I ran the script and it wasn't able to find tokenizer after it trained it , where do we write the trained tokenizer to on disk ?

PR #402, opened 2023-06-09
Update targets and runtime budgets.

fsschneider:
This PR updates the (test) targets and maximum runtime budgets according to Table 7 of the AlgoPerf paper.

### DIFF
--- a/algorithmic_efficiency/workloads/criteo1tb/workload.py
+++ b/algorithmic_efficiency/workloads/criteo1tb/workload.py
@@ -33,7 +33,7 @@ class BaseCriteo1TbDlrmSmallWorkload(spec.Workload):
 
   @property
   def test_target_value(self) -> float:
-    return 0.126053
+    return 0.126060
 
   @property
   def loss_type(self) -> spec.LossType:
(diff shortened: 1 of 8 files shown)


priyakasimbeg (changes requested):
From the latest pytorch measurements it seems like some of the pytorch workloads are slower than some of the jax workloads. 
Do we want to set the maximum runtime to the maximum extrapolated time between the two frameworks? If so we'd want to increase criteo to ~3.5 hours, ogbg to 6.5 hours, and wmt to ~16 hours.
For reference the exact numbers are here in the [Time Comparison Jax vs Pytorch (deepspeech fix) tab](https://docs.google.com/spreadsheets/d/1PGuF-e_hux5pYHFawr0Q62mDzGNiNX3FuvlsN2MbsOQ/edit?resourcekey=0-Onr8x3KwAuXIWEcocaqKnQ#gid=804216215).
In any case we have updated deepspeech numbers ready too so we can integrate those in this PR to ~17 hours.

georgedahl:
I believe we decided in one of the working group meetings to just use the jax-derived time budgets.

znado (approved):

priyakasimbeg (approved):
As discussed in working group meeting, we'll stick to the jax times to set the maximum runtimes.

PR #405, opened 2023-06-20
Version updates & PyTorch upgrade

pomonam:
This PR performs various version updates (so that we have up-to-date dependencies). This includes an update to PyTorch 2.0, where we can have a flag `--torch_compile` to enable JIT compile. 

Some notes:
- I have only tested `torch.compile` on smaller workloads; it would be great if we could test whether it works for all the workloads.
- I have to port almost all Transformer codes (for WMT) from PyTorch version 1.13 due to changes in the API. It would be useful to check this workload extra carefully.

### DIFF
--- a/algorithmic_efficiency/workloads/wmt/wmt_pytorch/models.py
+++ b/algorithmic_efficiency/workloads/wmt/wmt_pytorch/models.py
@@ -217,10 +217,10 @@ class TransformerEncoder(nn.Module):
         rate is high.
 
   Examples::
-      >>> encoder_layer = nn.TransformerEncoderLayer(d_model=512, nhead=8)
-      >>> transformer_encoder = nn.TransformerEncoder(encoder_layer, num_layers=6)
-      >>> src = torch.rand(10, 32, 512)
-      >>> out = transformer_encoder(src)
+    >>> encoder_layer = nn.TransformerEncoderLayer(12, 8)
+    >>> transformer_encoder = nn.TransformerEncoder(encoder_layer, 6)
+    >>> src = torch.rand(10, 32, 512)
+    >>> out = transformer_encoder(src)
   """
   __constants__ = ['norm']
 
@@ -283,7 +283,8 @@ class TransformerEncoder(nn.Module):
         f' {str_first_layer}.activation_relu_or_gelu was not True'
     elif not first_layer.norm1.eps == first_layer.norm2.eps:
       why_not_sparsity_fast_path = \
-        f'{str_first_layer}.norm1.eps was not equal to {str_first_layer}.norm2.eps'
+        f'{str_first_layer}.norm1.eps was not equal ' \
+        f'to {str_first_layer}.norm2.eps'
     elif not src.dim() == 3:
       why_not_sparsity_fast_path = \
         f'input not batched; expected src.dim() of 3 but got {src.dim()}'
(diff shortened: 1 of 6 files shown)


algorithmic_efficiency/workloads/wmt/wmt_pytorch/models.py
```diff
@@ -209,6 +203,159 @@ def forward(self,
...
+      _skpm_dtype = src_key_padding_mask.dtype  # pylint: disable=invalid-name
+      if _skpm_dtype != torch.bool and not torch.is_floating_point(
+          src_key_padding_mask):
+        raise AssertionError(
+            'only bool and floating types of key_padding_mask are supported')
+    output = src
+    convert_to_nested = False
+    first_layer = self.layers[0]
+    src_key_padding_mask_for_layers = src_key_padding_mask
+    why_not_sparsity_fast_path = ''
```
runame, line 265: As mentioned in our meeting earlier, a lot of the code here can be simplified a lot since the conditions for using the fast path implementation will never be fulfilled in the workload anyway. But we can do this in a later clean-up PR.

algorithmic_efficiency/workloads/wmt/wmt_pytorch/models.py
```diff
@@ -431,23 +568,174 @@ def __init__(self,
...
+        see the docs in Transformer class.
+    """
+    if src_key_padding_mask is not None:
+      _skpm_dtype = src_key_padding_mask.dtype  # pylint: disable=invalid-name
+      if _skpm_dtype != torch.bool and not torch.is_floating_point(
+          src_key_padding_mask):
+        raise AssertionError(
+            'Only bool and floating types of key_padding_mask are supported')
+    # See Fig. 1 of https://arxiv.org/pdf/2002.04745v1.pdf.
+    why_not_sparsity_fast_path = ''
```
runame, line 613: Same as above.

algorithmic_efficiency/workloads/wmt/wmt_pytorch/models.py
```diff
@@ -1055,44 +1401,52 @@ def multi_head_attention_forward(
...
+      attn_output_weights = torch.bmm(q_scaled, k.transpose(-2, -1))
+
+    # Optionally average attention weights over heads.
     attn_output_weights = attn_output_weights.view(bsz,
                                                    num_heads,
                                                    tgt_len,
                                                    src_len)
     if average_attn_weights:
-      attn_output_weights = attn_output_weights.sum(dim=1) / num_heads
+      attn_output_weights = attn_output_weights.mean(dim=1) / num_heads
```
runame, line 1449: What was the motivation for the change here? If `dim=1` is the dimension over the heads and we switch to `mean` instead of `sum`, the division by `num_heads` should be removed or not?

runame (commented):
I specifically took a brief look at the port of the transformer code for the WMT workload -- looks good, but I think a lot of the code can be removed/simplified since the fast path implementation is never used here anyway. This can be also done in a follow-up PR though.

algorithmic_efficiency/workloads/wmt/wmt_pytorch/models.py
pomonam: My bad! I will fix this.

znado (approved):
Nice! Thanks so much for handling this. Before we merge we should run the jax vs pytorch equivalence tests with `--torch_compile` for WMT and post the results.

PR #407, opened 2023-06-21
Drop support for Python 3.7

runame:
Since some of the recently updated packages don't support Python `3.7` anymore (e.g. `optax`), it looks like we also have to bump the minimum required Python version to `3.8`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 
 1. Create new environment, e.g. via `conda` or `virtualenv`:
 
-   Python minimum requirement >= 3.7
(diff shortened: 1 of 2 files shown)


pomonam (approved):
LGTM

runame:
> I think we're OK in the [dockerfile](https://github.com/mlcommons/algorithmic-efficiency/blob/main/docker/Dockerfile), because apt-get install python3 will install the latest 3.x?

Yeah, I think so too. Also, we are using Python `3.9` for all GitHub workflows.

PR #408, opened 2023-06-21
Update Dockerfile

sourabh2k15:
updating entrypoint to submission runner as we're done with docker debugging


priyakasimbeg:
This will break torchrun compatibility and break all of the timing scripts for the paper.

PR #409, opened 2023-06-22
Add baseline tests to CI tests

priyakasimbeg:
Add test_baselines.py to CI workflow, for 12 enabled baselines this adds about 4:37 minutes to the CI tests.
Also fix bug in mnist input_pipeline.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -182,3 +182,5 @@ jobs:
         pytest -vx tests/test_param_shapes.py
         pytest -vx tests/test_param_types.py
         pytest -vx tests/test_ssim.py
+        pytest -vx tests/submission_runner_test.py
+        pytest -vx tests/test_baselines.py
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -182,5 +182,3 @@ jobs:
         pytest -vx tests/test_param_shapes.py
         pytest -vx tests/test_param_types.py
         pytest -vx tests/test_ssim.py
-        pytest -vx tests/submission_runner_test.py
-        pytest -vx tests/test_baselines.py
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -181,4 +181,6 @@ jobs:
         pytest -vx tests/test_num_params.py
         pytest -vx tests/test_param_shapes.py
         pytest -vx tests/test_param_types.py
-        pytest -vx tests/test_ssim.py
\ No newline at end of file
+        pytest -vx tests/test_ssim.py
+        pytest -vx tests/submission_runner_test.py
+        pytest -vx tests/test_baselines.py
```
znado, line 186: will this download the MNIST data? that'll be a bit slow but probably is OK

.github/workflows/CI.yml
```diff
@@ -181,4 +181,6 @@ jobs:
         pytest -vx tests/test_num_params.py
         pytest -vx tests/test_param_shapes.py
         pytest -vx tests/test_param_types.py
-        pytest -vx tests/test_ssim.py
\ No newline at end of file
+        pytest -vx tests/test_ssim.py
+        pytest -vx tests/submission_runner_test.py
```
znado, line 185: do we even need the `test_submission` function in this file anymore? it seems like that's covered by test_baselines.py now

znado (changes requested):

.github/workflows/CI.yml
priyakasimbeg: I suppose we don't I removed it.

.github/workflows/CI.yml
priyakasimbeg: It took about 4 minutes 37 seconds to run all 12 baseline tests I enabled so I think it is ok too.

znado (approved):

PR #411, opened 2023-06-22
Update criteo pytorch eval bsz

priyakasimbeg:
Reduce criteo eval bsz to avoid OOM on eval step.


runame (approved):
Maybe we should remove the `eval_batch_size` property from the workload parent class, to avoid confusion. The Jax version of the workload already has the property implemented anyway.

PR #417, opened 2023-06-26
Dockerfile changes and regression tests

priyakasimbeg:
Dockerfile changes:
- Add arg to Dockerfile to specify whether to install jax gpu or pytorch gpu in image.
- Add option to run in internal mode for syncing w our internal GCP bucket

Regression tests:
- Add regression tests with self hosted runner for 8 workloads on adamw in both pytorch and jax.

Fixes to CI test for baselines
- remove itertools.cycle on mnist dataset. This increases the memory usage and causes the github hosted runners (7GB in RAM to OOM).


docker/Dockerfile
```diff
@@ -21,12 +28,31 @@ RUN mkdir -p experiment_runs/
...
+        echo "Installing Jax GPU" \
+        && cd algorithmic-efficiency \
+        && pip install -e '.[jax_gpu]' -f 'https://storage.googleapis.com/jax-releases/jax_cuda_releases.html' \
+        && pip install -e '.[pytorch_cpu]' -f 'https://download.pytorch.org/whl/torch_stable.html'; \
+    elif [ "$framework" = "pytorch" ] ; then \
+        echo "Installing Pytorch GPU" \
+        && cd algorithmic-efficiency \
+        && pip install -e '.[jax_cpu]' \
+        && pip install -e '.[pytorch_gpu]' -f 'https://download.pytorch.org/whl/torch_stable.html'; \
+    else \
```
znado, line 49: maybe instead of this we should have "the default be "both", and this case is for "both" and the else case prints an error and exits

docker/build_docker_images.sh
```diff
@@ -0,0 +1,19 @@
+for FRAMEWORK in "jax" "pytorch" "joint"
```
znado, line 1: nit: can you rename this to build_all_dev_docker_images.sh?

znado, line 1: should "joint" be "both" here?

.github/workflows/regression_tests.yml
```diff
@@ -0,0 +1,117 @@
...
+on: [push, pull_request]
+
+jobs:
+  fastmri_jax:
+    runs-on: self-hosted
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run containerized workload
+      run: |
+        docker run  -v /home/kasimbeg/data/:/data/ -v /home/kasimbeg/experiment_runs/:/experiment_runs -v /home/kasimbeg/experiment_runs/logs:/logs --gpus all --ipc=host us-central1-docker.pkg.dev/training-algorithms-external/mlcommons-docker-repo/algoperf_jax:dev -d fastmri -f jax -s baselines/adamw/jax/submission.py -w fastmri -t baselines/adamw/tuning_search_space.json -e test_today/adamw -m 10 -c False -o True -r false
```
znado, line 12: no big deal if not, but is it possible to pull `/home/kasimbeg` out into a variable?

znado (dismissed):
thanks for putting this together!

.github/workflows/regression_tests.yml
priyakasimbeg: Yes, that is a good suggestion. I can just make it $HOME for now?

docker/build_docker_images.sh
priyakasimbeg: Sorry, I hadn't finalized this yet. I added a flag to specify the branch. Is that ok w you?

priyakasimbeg: yes, fixed

docker/Dockerfile
priyakasimbeg: good idea

.github/workflows/regression_tests.yml
znado: sgtm

znado (approved):

priyakasimbeg:
@runame could you check that the setup.cfg and startup.sh here are installing the pytorch versions consistent with the fix you mentioned for the CUDNN libraries?

runame (commented):
`setup.cfg` looks good, not sure how `startup.sh` influences which cudnn version is used?

PR #418, opened 2023-06-27
Update docker README.md

priyakasimbeg:
Fix args in docker documentation for criteo. Should be criteo1tb.
Also check for valid args in startup.sh.


runame (approved):
Thanks for the fix!

znado (approved):

PR #419, opened 2023-06-27
Update pytorch version in setup.cfg

priyakasimbeg:


runame (approved):
This will definitely work, have not checked if it breaks for `2.0.1+cu117`, but it probably will since it doesn't work for `1.13.0+cu117`. Also, Jax ships with CUDA 11.8 wheels anyway.

znado (approved):

PR #420, opened 2023-06-28
Fix `startup.sh`

runame:
Simple syntax fixes -- please check if this is what was intended.


priyakasimbeg (approved):
Yes, thanks for catching that!

PR #427, opened 2023-06-30
Documentation, scoring and dataset updates 

priyakasimbeg:
Documentation updates:
- Update README with Docker installation instructions
- Update getting started documentation with Docker instructions
- Added instructions to CONTRIBUTING.md relating to Docker workflows and testing.

Scoring:
- Add e2e scoring script usable for external folks

Imagenet data setup fix:
- Fix imagenet dataset: add tfds.builder and download_and_prepare to extract and process the imagenet dataset into tfrecords for jax workloads.

Other:
- make docker entrypoint script return exit code 1 if submission runner throws an error.


priyakasimbeg:
@pomonam @chandramouli-sastry this PR contains updated Docker instructions. See "Docker workflows" under CONTRIBUTING.md.
Link for easier reading and navigating: https://github.com/mlcommons/algorithmic-efficiency/blob/metrics_logging/CONTRIBUTING.md#docker-workflows
Let me know if you have any suggestions.

scoring/score_submission.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def main(_):
+  df = scoring_utils.get_experiment_df(FLAGS.experiment_path)
+  results = {
+      FLAGS.submission_tag: df,
+  }
+  performance_profile_df = scoring.compute_performance_profiles(
+      results,
+      workload_metadata,
```
znado, line 180: ideally `compute_performance_profiles` would be able to be called from a script (like if a user is using it in a pipeline), and it is a pain to have to supply `workload_metadata` organized like this. ideally we could just look up that metadata given the workload name. this would be possible today but it's annoyingly not because we don't have the metric names in the spec api, just `has_reached_test_target`. I think we should just add

```python
def target_metric_name(self):
  return 'name'
```

to each workload, and then inside `scoring.py:get_index_that_reaches_target(...)` we look up all the info we need from the workload object. I did this for us in [[this PR]](https://github.com/mlcommons/algorithmic-efficiency/pull/443), if you merge that into your branch then I think this will be much cleaner.

datasets/dataset_setup.py
```diff
@@ -55,7 +55,7 @@
 Example command:
 
 python3 datasets/dataset_setup.py \
-  --data_dir=~/data \
+  --data_dir=~/data/imagenet \
```
znado, line 58: undo?

datasets/dataset_setup.py
```diff
@@ -367,15 +368,29 @@ def setup_imagenet_jax(data_dir):
...
+  if not os.path.exists(
+      os.path.join(manual_download_dir, IMAGENET_TRAIN_TAR_FILENAME)):
+    logging.info('Copying {} to {}'.format(train_tar_file_path,
+                                           manual_download_dir))
+    shutil.copy(train_tar_file_path, manual_download_dir)
+  if not os.path.exists(
+      os.path.join(manual_download_dir, IMAGENET_VAL_TAR_FILENAME)):
+    logging.info('Copying {} to {}'.format(val_tar_file_path,
+                                           manual_download_dir))
+    shutil.copy(val_tar_file_path, manual_download_dir)
```
znado, line 387: can we avoid having to do these copies? they could take up a lot of disk space (at least the train tar) if we have 2 copies

algorithmic_efficiency/logger_utils.py
```diff
@@ -266,11 +284,13 @@ class MetricLogger(object):
 
   def __init__(self,
                csv_path: str = '',
+               eval_csv_path: str = '',
```
znado, line 287: if these are required we should probably not default them

CONTRIBUTING.md
```diff
@@ -7,3 +7,225 @@ Generally we encourage people to become a MLCommons member if they wish to contr
...
+
+# Setup 
+## Setting up a Linux VM on GCP
+If you want to run containers on GCP VMs or store and retrieve Docker images from the Google Cloud Container Registry, please read ahead.
+If you'd like to use a Linux VM, you will have to install the correct GPU drivers and the NVIDIA Docker toolkit.
+We recommmend to use the Deep Learning on Linux image. Further instructions are based on that.
+
+### Installing GPU Drivers
+You can use the `scripts/cloud-startup.sh` as a startup script for the VM. This will automate the installation of the
+NVIDIA GPU Drivers and NVIDIA Docker toolkit.
```
znado, line 27: undo line break?

znado (changes requested):

datasets/dataset_setup.py
priyakasimbeg: No this is intentional. Unless we want to automate naming the directories for a specific dataset?

scoring/score_submission.py
```diff
@@ -6,178 +6,73 @@
...
+
 def main(_):
   df = scoring_utils.get_experiment_df(FLAGS.experiment_path)
   results = {
       FLAGS.submission_tag: df,
   }
   performance_profile_df = scoring.compute_performance_profiles(
       results,
-      workload_metadata,
+    #   workload_metadata,
```
znado, line 75: delete

scoring/score_submission.py
```diff
@@ -6,178 +6,73 @@
 import scoring_utils
 
 from algorithmic_efficiency import workloads
```
znado, line 8: we can delete this import and the `WORKLOADS` var below right?

scoring/scoring.py
```diff
@@ -41,11 +49,102 @@
     'wer',
     'l1_loss',
 ]
+
 MAX_EVAL_METRICS = ['average_precision', 'ssim', 'accuracy', 'bleu_score']
+WORKLOADS = {
```
znado, line 54: we should be able to just import workloads and use the var/import helper from there, right?

scoring/scoring_utils.py
```diff
@@ -174,11 +173,13 @@ def get_experiment_df(experiment_dir):
       try:
         trial_df = pd.read_csv(eval_measurements_filepath)
       except FileNotFoundError as e:
+        logging.info(f"Could not read {eval_measurements_filepath}")
```
znado, line 176: nit: single quotes

scoring/scoring.py
```diff
@@ -128,7 +227,13 @@ def get_index_that_reaches_target(workload_df,
       lambda x: op(x, validation_target))
   test_target_reached = test_series.apply(lambda x: op(x, test_target))
 
-  target_reached = validation_target_reached & test_target_reached
+  print(validation_target_reached)
```
znado, line 230: delete prints, comments (here and below)

znado (dismissed):
looks great, thanks for wiring all this up!!

scoring/scoring_utils.py
```diff
@@ -102,7 +103,7 @@ def get_trials_df_dict(logfile):
     """
   trials_dict = get_trials_dict(logfile)
   trials_df_dict = {}
-  for trial in trials_dict:
+  for trial in trials_dict.items():
```
znado, line 106: is this supposed to be items() or keys()?

setup.cfg
```diff
@@ -44,6 +44,7 @@ install_requires =
   gputil==1.4.0
   psutil==5.9.5
   clu==0.0.7
+  matplotlib==3.7.2
```
znado, line 47: nit: can you make it matplotlib>=3.7.2 because we really don't care if it's the exact version or not because all this does is plot?

znado (approved):

scoring/scoring_utils.py
priyakasimbeg: yup, oops that is supposed to be keys(). Thanks for catching that.

setup.cfg
priyakasimbeg: fixed

PR #430, opened 2023-07-06
Fix `_get_time`

runame:
Unless I'm missing something we always need to call `torch.cuda.synchronize()` for timing, not only when DDP is used. Also, I think this shouldn't have unwanted side-effects for the Jax runs.

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -176,7 +176,8 @@ USE_PYTORCH_DDP, RANK, DEVICE, N_GPUS = pytorch_setup()
 
 
 def _get_time():
-  torch.cuda.synchronize()
+  if torch.cuda.is_available():
+    torch.cuda.synchronize()
   return time.time()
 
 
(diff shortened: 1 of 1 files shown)


priyakasimbeg (approved):

priyakasimbeg:
Just noticed the baseline tests on CPU fail, because torch.cuda throws an error when run in CPU mode?
Can we check if we're using the GPU (for both pytorch and jax) before calling synchronize?

priyakasimbeg (changes requested):
Baseline tests on CPU fail when calling torch.cuda.synchronize. I think we should add a check whether we're running on cpu before calling torch.cuda.synchronize. Also torch may not be installed if submitters are using just jax framework, so we should add a check for the framework as well.

runame:
Sorry, fixed! `torch` should always be installed, even when someone only wants to run `jax` workloads.

priyakasimbeg (approved):
LGTM! Thanks for the fix!

znado (approved):

PR #433, opened 2023-07-06
Fix criteo1tb eval batch size for PyTorch

runame:
Set PyTorch eval batch size equal to training batch size and remove property from workload parent class to avoid confusion.


priyakasimbeg (approved):

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/workload.py
```diff
@@ -22,7 +22,7 @@ class Criteo1TbDlrmSmallWorkload(BaseCriteo1TbDlrmSmallWorkload):
 
   @property
   def eval_batch_size(self) -> int:
-    return 65_536
+    return 262_144
```
znado, line 25: did you try 524_288 (the size that jax uses)?

znado (approved):

PR #438, opened 2023-07-13
Revive checkpointing for Pytorch DDP

### LINKED ISSUE #412: Resuming from checkpoint fails when using TORCH_DDP
## Description

Resuming from a checkpoint using --resume_last_run when using multiple threads with `torchrun` has two issues:

First, the `DistributedDataParallel` is not unwrapped, so loading the model parameters fails, meaning you effectively start from a newly initialized model.

Second, if that is fixed, training hangs immediately after the start of the training loop.

## Steps to Reproduce

`torchrun --nnodes=1 --nproc_per_node=8 submission_runner.py --framework=pytorch --resume_last_run ...`

## Source or Possible Fix

For the first problem, in `algorithmic_efficiency/checkpoint_utils.py`, replace both instances of `if isinstance(model_params, torch.nn.DataParallel)` with `hasattr(model_params, 'module')` to correctly unwrap DistributedDataParallel.

For the second problem, `get_log_dir()` in `algorithmic_efficiency/logger_utils.py` returns None for all processes except rank 0, so rank 0 loads the checkpoint and the remaining processes don't, which causes the processes to have inconsistent states and hang during an all_wait. I have a local band-aid fix but this needs attention from someone who understands the existing code.

pomonam:
Aiming to fix #412.


znado (approved):

znado:
In the GH UI I rebased it to dev instead of main and that introduced some issues (sorry!), can you try rebasing your local branch to dev and syncing and reuploading?

priyakasimbeg:
This PR is stale, will address this issue in new PR.

PR #439, opened 2023-07-16
Attention param type test fix

znado:


priyakasimbeg (changes requested):
There is one CI test failing still: 
```
__________________ ERROR collecting tests/test_param_types.py __________________
In test_param_types: function uses no argument 'workload'
```

znado:
> There is one CI test failing still:
> 
> ```
> __________________ ERROR collecting tests/test_param_types.py __________________
> In test_param_types: function uses no argument 'workload'
> ```

fixed

Issue #426, opened 2023-06-30
Pytorch Conformer CUDA error


priyakasimbeg:
Pytorch Conformer OOMs on GPU with adamw. 
branch: dev
test link: https://github.com/mlcommons/algorithmic-efficiency/actions/runs/5416731116/jobs/9846849204 
Expand librispeech_conformer_pytorch and then expand "Run containerized workload".

## Description
Traceback:
```
2m 48s
  File "/algorithmic-efficiency/baselines/adamw/pytorch/submission.py", line 98, in update_params
    loss.backward()
  File "/usr/local/lib/python3.8/dist-packages/torch/_tensor.py", line 487, in backward
    torch.autograd.backward(
  File "/usr/local/lib/python3.8/dist-packages/torch/autograd/__init__.py", line 200, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 2.51 GiB (GPU 0; 15.78 GiB total capacity; 6.25 GiB already allocated; 283.44 MiB free; 13.91 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
Traceback (most recent call last):
  File "submission_runner.py", line 681, in <module>
    app.run(main)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
  File "submission_runner.py", line 652, in main
    score = score_submission_on_workload(
  File "submission_runner.py", line 577, in score_submission_on_workload
    timing, metrics = train_once(workload, global_batch_size,
  File "submission_runner.py", line 365, in train_once
    optimizer_state, model_params, model_state = update_params(
  File "/algorithmic-efficiency/baselines/adamw/pytorch/submission.py", line 98, in update_params
    loss.backward()
  File "/usr/local/lib/python3.8/dist-packages/torch/_tensor.py", line 487, in backward
    torch.autograd.backward(
  File "/usr/local/lib/python3.8/dist-packages/torch/autograd/__init__.py", line 200, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[...]

PR #442, opened 2023-07-20
Update submission_runner.py to prevent torch Conformer OOM

sourabh2k15:
This is to fix the OOM error seen in https://github.com/mlcommons/algorithmic-efficiency/issues/426

We use torch.cuda.empty_cache() after evals so it doesn't run out of memory.

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -453,6 +453,8 @@ def train_once(
                   save_intermediate_checkpoints=FLAGS
                   .save_intermediate_checkpoints)
 
+          if USE_PYTORCH_DDP:
+            torch.cuda.empty_cache()
           logging_end_time = get_time()
 
           train_state['accumulated_logging_time'] += (


pomonam:
Nice! Could this also solve the CUDA OOM issue for the Criteo workloads?

sourabh2k15:
@pomonam if it's an eval related OOM then yes ... will check w/ @priyakasimbeg

runame:
Could potentially be eval related since it always fails during the first update step after the first eval (and not during the first update step before the first eval).

submission_runner.py
```diff
@@ -453,6 +453,8 @@ def train_once(
                   save_intermediate_checkpoints=FLAGS
                   .save_intermediate_checkpoints)
 
+          if USE_PYTORCH_DDP:
```
runame, line 456: Should this really only be called when PyTorch DDP is used or also when just DataParallel or a single GPU is used?

sourabh2k15: hmmm that's a good question, if it's OOMing in the other cases (DataParallel or single GPU) we should use the condition you suggest 

right now I actually don't know if it OOMs in those other cases, maybe we just go ahead and do it for all cases to be safe ?

sourabh2k15: okay I'm gonna merge this for now and test OOM for other cases, will send a fix

runame: Yeah let's just do it for all cases, I don't see any downside to it.

PR #443, opened 2023-07-20
adding `def target_metric_name(self) -> str:` property

znado:


priyakasimbeg (approved):
Thanks this will make things much cleaner!

Issue #425, opened 2023-06-30
Pytorch Criteo CUDA error


priyakasimbeg:
Branch: dev
Test link: https://github.com/mlcommons/algorithmic-efficiency/actions/runs/5416731116/jobs/9846848568 
For details expand criteo_pytorch and then expand Run Containerized Workload.

## Description
Criteo Pytorch OOMs.

Traceback:
```
ret = func(self, *args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/torch/optim/adamw.py", line 171, in step
    adamw(
  File "/usr/local/lib/python3.8/dist-packages/torch/optim/adamw.py", line 321, in adamw
    func(
  File "/usr/local/lib/python3.8/dist-packages/torch/optim/adamw.py", line 566, in _multi_tensor_adamw
    denom = torch._foreach_add(exp_avg_sq_sqrt, eps)
torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 2.00 GiB (GPU 3; 15.78 GiB total capacity; 12.14 GiB already allocated; 307.44 MiB free; 14.14 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
```

## Steps to Reproduce
On kasimbeg-3
```
docker pull us-central1-docker.pkg.dev/training-algorithms-external/mlcommons-docker-repo/algoperf_jax_dev 
```
```
  docker run  -v $HOME/data/:/data/ -v $HOME/experiment_runs/:/experiment_runs -v $HOME/experiment_runs/logs:/logs --gpus all --ipc=host us-central1-docker.pkg.dev/training-algorithms-external/mlcommons-docker-repo/algoperf_pytorch_dev -d criteo1tb -f pytorch -s baselines/adamw/pytorch/submission.py -w criteo1tb -t baselines/adamw/tuning_search_space.json -e test_today/adamw -m 10 -c False -o True -r false
```

priyakasimbeg:
Update, I made sure I ran with --torch_compile flag and now the submission_runner crashes with `CUDA error:misaligned address`. 
Full trace back in  https://github.com/mlcommons/algorithmic-efficiency/actions/runs/5534559953/jobs/10099946892 under criteo_pytorch > Run containerized workload.

runame:
[...]

PR #445, opened 2023-07-23
Add options to torch.compile

pomonam:
Some workloads cause issues when enabling `torch.compile`. This PR implements several if-conditions so that workloads that cause issues for `torch.compile` get automatically disabled (or compiled with a different backend).

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -295,14 +295,14 @@ def train_once(
         logging.warning(
             'These workloads cannot be fully compiled under current '
             'PyTorch version. Proceeding with `backend=eager`.')
-        model_params = torch.compile(model_params, backend="eager")
+        model_params = torch.compile(model_params, backend='eager')
       elif FLAGS.workload in aot_eager_backend_workloads:
         logging.warning(
             'These workloads cannot be fully compiled under current '
             'PyTorch version. Proceeding with `backend=aot_eager`.')
-        model_params = torch.compile(model_params, backend="aot_eager")
+        model_params = torch.compile(model_params, backend='aot_eager')
       else:
-        logging.warning('Performing `torch.compile`.')
+        logging.info('Performing `torch.compile`.')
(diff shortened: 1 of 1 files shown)


submission_runner.py
```diff
@@ -282,7 +282,29 @@ def train_once(
...
+      aot_eager_backend_workloads = ['criteo1tb']
+      if FLAGS.workload in compile_error_workloads:
+        logging.warning(
+            'These workloads cannot be fully compiled under current '
+            'PyTorch version. Proceeding without `torch.compile`.')
+      elif FLAGS.workload in eager_backend_workloads:
+        logging.warning(
+            'These workloads cannot be fully compiled under current '
+            'PyTorch version. Proceeding with `backend=eager`.')
+        model_params = torch.compile(model_params, backend="eager")
```
znado, line 298: nit: single quotes for 'eager' and 'aot_eager'

submission_runner.py
```diff
@@ -282,7 +282,29 @@ def train_once(
...
+      elif FLAGS.workload in aot_eager_backend_workloads:
+        logging.warning(
+            'These workloads cannot be fully compiled under current '
+            'PyTorch version. Proceeding with `backend=aot_eager`.')
+        model_params = torch.compile(model_params, backend="aot_eager")
+      else:
+        logging.warning('Performing `torch.compile`.')
```
znado, line 305: I would make this logging.info

znado (approved):
LGTM! does this allow all of the workloads in pytorch to run now with the latest pytorch version?

submission_runner.py
pomonam: Thank you!

pomonam:
Thank you for the review! Yes, now should now run except for compile error (except for the Criteo workload which still gives OOM; #425).

znado (approved):

PR #460, opened 2023-07-27
documentation updates

priyakasimbeg:


README.md
```diff
@@ -113,6 +112,7 @@ See instructions [here](https://github.com/NVIDIA/nvidia-docker).
 
 
 ### Running Docker Container (Interactive)
+To use the Docker container as an interactive virtual environment, you can run a container mounted to your local data and code directories and execute the `bash` program. This may be useful if you are in the process of developing a submission.
```
znado, line 115: would it make more sense to rewrite this as:

"To use the Docker container as an interactive virtual environment, you can run a container with your local data and code directories mounted and execute the `bash` program. This may be useful if you are in the process of developing a submission."

CONTRIBUTING.md
```diff
@@ -172,7 +172,7 @@ docker run -t -d \
 --gpus all \
 --ipc=host \
 <docker_image_name> \
--b <debug_mode>
+-b true
```
znado, line 175: should this be hardcoded true (here and below)?

znado (approved):

CONTRIBUTING.md
priyakasimbeg: yes this is intentional for contributors who will probably want the container to persist in most cases.

README.md
priyakasimbeg: yes that is easier to read. thanks!

PR #465, opened 2023-07-28
Fixing librispeech dataset setup scripts

sourabh2k15:
1) Fixed tar extraction paths for librispeech split tar.gz files that are downloaded.
2) Fixed tokenizer write path so it's written to the same directory that's provided.
3) Fixed final output directory path in preprocessing script so records are written to the data directory provided.
4) Tokenizer will be trained if there's no existing tokenizer thus eliminating the need for train_tokenizer flag

With these changes one should be able to run librispeech setup e2e by running :
```
python3 datasets/dataset_setup.py --data_dir=/data/ --temp_dir=/data/ --librispeech 
```

This will result in final data being written in `/data/librispeech_preprocessed/*` and raw extracted data being written in `/data/librispeech/Librispeech`, vocab file will be written to `/data/librispeech/Librispeech/spm_model.vocab`


priyakasimbeg (approved):
LGTM! Thank you for fixing this dataset!

PR #484, opened 2023-08-14
PyTorch Workload Fixes

pomonam:
This PR fixes several issues with PyTorch workloads. 

1. Criteo OOM issues are fixed on the nightlies (#425). However, it still shows an OOM error on the current version (2.0.1). 
2. WMT workload now works with `torch.compile` (both nightly and 2.0.1). It slightly improves the timing results.
3. Deepspeech workload now works with `torch.compile` (both nightly and 2.0.1). It slightly improves the timing results.
4. Minor style fixes.

Note: OGBG also runs with `torch.compile` on nightly but somehow slows down the training. The previous issue was that we used jgraph as inputs (instead of Tensors). So I am keeping its default behavior to be "not compiled".

### DIFF
--- a/algorithmic_efficiency/workloads/criteo1tb/criteo1tb_jax/workload.py
+++ b/algorithmic_efficiency/workloads/criteo1tb/criteo1tb_jax/workload.py
@@ -1,4 +1,5 @@
 """Criteo1TB workload implemented in Jax."""
+
 import functools
 from typing import Dict, Optional, Tuple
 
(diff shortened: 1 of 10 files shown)


priyakasimbeg (approved):
Hi Juhan, 
That is great news! Thank you for fixing the torch compile issues with deepspeech and WMT and testing the nightly pytorch on Criteo!
I'll merge these changes in now and we can discuss our options tomorrow as to what Pytorch version to pin.

Issue #467, opened 2023-08-08
WMT slower in Pytorch than Jax


priyakasimbeg:
WMT pytorch is currently slower than Jax.

This bug is intended to at least document possible causes. 

@runame @pomonam could you please summarize current findings and possible solutions?

msaroufim:
FWIW I did try using the HF implementation with boolean masks and that `torch.compile`'d just fine https://gist.github.com/msaroufim/946d7d26e89bab0bfe83d9929b533701

msaroufim:
@runame I managed to `torch.compile()` the wmt model just now an A10G using pytorch nightlies

`pip3 install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu118 --force-reinstall`

I don't have perf numbers yet since it feels stuck at `Translating evaluation dataset` - do you mind trying it out on your end as well and lmk if you hit any new errors?

pomonam:
@msaroufim Thank you so much for looking into this! Yes, I will try this and see if there are any new issues.

runame:
@msaroufim @pomonam Thanks a lot for investigating! I won't get to working on this before Sunday or Monday, will check this thread for updates then.

pomonam:
@runame One thing I wanted to try out (but did not have time to do) was to use the PyTorch default functions (e.g., `multi_head_attention_forward`) when don't need cache - training. I am not sure how useful this might be, but I think it is worth a try. Also, with `torch.compile`, the graph seems to break many times (and the timings don't change much), but I am not sure how to fix this.

msaroufim:
Oh interesting so you can confirm that it does compile but you're getting graph breaks? Are you running `fullgraph=True`? 

Also could you print the logs you're getting

runame:
Branch `dev` (also with changes from #489, also with `torch.compile` before `DDP`) + stable + `fullgraph=True`:

```
Traceback (most recent call last):
  File "submission_runner.py", line 624, in <module>
    app.run(main)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
  File "submission_runner.py", line 595, in main
    score = score_submission_on_workload(
  File "submission_runner.py", line 520, in score_submission_on_workload
    timing, metrics = train_once(workload, global_batch_size,
[...]

PR #489, opened 2023-08-17
Refactor of `MultiheadAttention` module in PyTorch WMT workload

runame:
This is supposed to help with #467. I refactored the `MultiheadAttention` module to remove a lot of unnecessary code and fused the parameter matrices for the linear input projection, which leads to a slight improvement in speed.

**Timing results**
(note that all timings but the one of this refactor alone are just run once and there is probably a significant variance over multiple runs)

Original score (Juhan's timing): 4379
Score with this refactor: 4196/4253/4265
Score with `torch.compile` (Juhan's timing): 4236
Score with this refactor and `torch.compile`: 4163
Score with `torch.compile` and nightly (Juhan's timing): 4161
Score with this refactor, `torch.compile`, and nightly: error -> the last validation set batch size is different which causes this error. Can easily be fixed but since we are going to stick to stable for now I would defer making any changes until we decide to upgrade.

Clearly, all improvements are very minor and are not enough to bridge the gap to Jax.

@pomonam Can you please review my refactor of the `MultiheadAttention` module?
@znado Can you please review my changes to the `test_param_types` test?

*Note: This breaks the traindiffs code for WMT, will be fixed in a follow up PR.*

### DIFF
--- a/algorithmic_efficiency/param_utils.py
+++ b/algorithmic_efficiency/param_utils.py
@@ -41,6 +41,10 @@ def pytorch_param_types(
     elif 'attn' in name or 'attention' in name:
       if 'bias' in name:
         param_types[name] = spec.ParameterType.ATTENTION_BIAS
+      elif 'in_proj' in name:
+        param_types[name] = spec.ParameterType.ATTENTION_QKV
+      elif 'kv_proj' in name:
+        param_types[name] = spec.ParameterType.ATTENTION_KV
       elif 'k_proj' in name or 'key' in name:
         param_types[name] = spec.ParameterType.ATTENTION_K
       elif 'q_proj' in name or 'query' in name:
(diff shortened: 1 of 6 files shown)


pomonam (dismissed):
LGTM - thank you for fixing this! (It seems like every test is working now?)

runame:
> It seems like every test is working now?

None of the GitHub tests were broken, the traindiffs test has to be run manually. @chandramouli-sastry already has a patch ready for after this PR is merged.

runame:
I have added `torch.cuda.synchronize()` to the profiler (see latest commit). @pomonam Is there any reason not to do this?

pomonam:
@runame Thank you! No, it was an oversight.

znado (approved):

PR #502, opened 2023-08-24
Fix for Criteo OOM

pomonam:
I rewrote the code for `dot interaction` and `embedding`. These changes fix the OOM issue during training, but not the OOM issue during evaluation (which is quite strange) I further reduced the batch size during evaluation to fix this. This PR makes the Criteo workload on PyTorch runnable, but further investigation would be needed to make the eval faster.

### DIFF
--- a/algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/models.py
+++ b/algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/models.py
@@ -5,6 +5,10 @@ import math
 import torch
 from torch import nn
 
+from algorithmic_efficiency.pytorch_utils import pytorch_setup
+
+USE_PYTORCH_DDP, RANK, DEVICE, N_GPUS = pytorch_setup()
+
 
 class DotInteract(nn.Module):
   """Performs feature interaction operation between dense or sparse features."""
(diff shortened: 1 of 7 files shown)


priyakasimbeg:
Hi Juhan! This looks great, thank you sm for these fixes.
It seems like one of the tests is failing on the param shapes between jax and pytorch. It seems like this is intended though. @znado I'm not sure if we want to disable the test for Criteo?

runame:
I haven't looked at this yet, but if @pomonam has changed the embedding implementation the param shapes test has to be adjusted so that the shape is matched correctly to the Jax implementation, similar to the changes I had to make in #489 to adjust for the new `MultiheadAttention` implementation.

pomonam:
Thank you! I fixed the tests.

priyakasimbeg (approved):
LGTM

priyakasimbeg:
I LGTMd this but will wait for @runame merge this in after you've had a chance to review it. 
Just one question, we initialize the embedding layer from uniform distribution, do you know by any chance if that's consistent with how it was initialized before this PR @pomonam? I noticed the _reset_parameters uses xavier_uniform or normal distributions for initializations.

pomonam:
Yes, I think the initialization is identical to what was done before.

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/models.py
```diff
@@ -62,13 +52,21 @@ def __init__(self,
...
+    assert vocab_size % num_chucks == 0
+    self.embedding_table_chucks = []
     scale = 1.0 / torch.sqrt(self.vocab_size)
-    self.embedding_table.weight.data = scale * self.embedding_table.weight.data
+    for i in range(num_chucks):
+      chunk = nn.Parameter(
+          torch.Tensor(self.vocab_size // num_chucks, self.embed_dim))
+      chunk.data.uniform_(0, 1)
+      chunk.data = scale * chunk.data
+      self.register_parameter(f"embedding_chunk_{i}", chunk)
```
runame, line 67: Nit: single quotation marks.

runame (commented):
LGTM!

Just to make sure we should check for equivalence using the traindiffs test, but it might currently be broken due to the new embedding implementation. @chandramouli-sastry might be able to help with this.

pomonam:
Yes, diff test would be great!

pomonam:
Yes, diff test would be great!

algorithmic_efficiency/workloads/criteo1tb/criteo1tb_pytorch/models.py
pomonam: Ah, thank you for the catch :). Fixed!

priyakasimbeg (approved):

Issue #503, opened 2023-08-30
submission_runner.py hparam start and end index flags


znado:
`submission_runner.py` should take `--hparam-start-index` and `--hparam-end-index` that slice the hparam iterator. this will be useful for running hparam sweeps in parallel

PR #509, opened 2023-09-13
Hparam trial indices

priyakasimbeg:
Add support for hparam start and end indices to select subset of tuning trials when running submission_runner.py as described in https://github.com/mlcommons/algorithmic-efficiency/issues/503.


runame (dismissed):
LGTM! Has to be merged with `dev` first though.

znado (dismissed):

Issue #279, opened 2023-01-04
Add a `--seed` flag to `submission_runner.py` to allow people to fix a seed


znado:
This obviously will not be used for any scoring runs, but will be useful for developing algos.

priyakasimbeg:
Fixed in https://github.com/mlcommons/algorithmic-efficiency/pull/514

PR #514, opened 2023-09-19
Add RNG seed flag to submission runner

priyakasimbeg:
For https://github.com/mlcommons/algorithmic-efficiency/issues/279. 

Add rng_seed flag to submission runner.
Save rng_seed in metadata.json file.

Also pin ml_dtypes version as temporary work around for jax dependency issue https://github.com/google/jax/issues/17693.

### DIFF
--- a/algorithmic_efficiency/logger_utils.py
+++ b/algorithmic_efficiency/logger_utils.py
@@ -275,8 +275,8 @@ def get_meta_data(workload: spec.Workload) -> dict:
   return meta_data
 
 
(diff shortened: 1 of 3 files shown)


znado (changes requested):
we also need to override rng_seed on L557 in the self tuning rule set

priyakasimbeg:
> we also need to override rng_seed on L557 in the self tuning rule set

@znado good catch. Fixed!

PR #515, opened 2023-09-19
Fix `shard_and_maybe_pad_np` function

runame:
This PR
- removes superfluous code in `pad` function,
- always pads to `global_batch_size` if it is provided -- even if there is no remainder,
- and fix how `pad_size` is used in `pad` function.


priyakasimbeg (approved):

priyakasimbeg:
Thanks for finding and fixing this issue

Issue #513, opened 2023-09-18
Librispeech Deepspeech pytorch training diverges


priyakasimbeg:
Librispeech Conformer training loss and WER increase after ~6000 steps.

## Description
<img width="700" alt="image" src="https://github.com/mlcommons/algorithmic-efficiency/assets/12614254/1b7fb1f0-9038-4ea1-9600-25a30264fd22">

## Steps to Reproduce
Git commit: 4c38ffb45baab8d77596f694a1e36d47d36a0448

Ran the following command to produce the target setting configuration:

```
torchrun --redirects 1:0,2:0,3:0,4:0,5:0,6:0,7:0 --standalone --nnodes=1 --nproc_per_node=8 submission_runner.py --framework=pytorch --workload=librispeech_deepspeech --submission_path=reference_algorithms/target_setting_algorithms/pytorch_nadamw.py --tuning_search_space=reference_algorithms/target_setting_algorithms/librispeech_deepspeech/tuning_search_space.json --data_dir=/data/librispeech --num_tuning_trials=1 --experiment_dir=/experiment_runs --experiment_name=targets_check_pytorch/nadamw_run_0 --overwrite=true --save_checkpoints=false --max_global_steps=36000 --librispeech_tokenizer_vocab_path=/data/librispeech/spm_model.vocab --torch_compile=true 2>&1 | tee -a /logs/librispeech_deepspeech_pytorch_<date_time>.log
```

priyakasimbeg:
From offline conversation w @sourabh2k15: both conformer and deepspeech can do that [... ] even internal 20 run experiments have trials that diverge.

I'll rerun with different seed.

priyakasimbeg:
Update: I reran this again and it did not diverge. However, it does not hit the new targets. 
One possible cause is that we did not update the warmup_steps in the target_setting algorithm hparams. I'll make sure we have the right hparams checked in and run this again.

priyakasimbeg:
The root cause of this is that the DeepspeechJax and DeepspeechPytorch workload classes are inheriting from the Conformer workload class but the step_hint and target properties are overridden in a generic Deepspeech class. Will send out PR shortly.

PR #526, opened 2023-10-02
Speech targets fix

priyakasimbeg:
Addresses issues in https://github.com/mlcommons/algorithmic-efficiency/issues/513.

- Override step_hints and targets in Deepspeech jax and pytorch workload classes. Without this fix the conformer properties are inherited.
- Remove Deepspeech base workload class. This is misleading since the jax and pytorch workload classes inherit from the respective conformer workload classes, so the overrides in this class are not being used.
- Fix hparams for target setting for speech workloads with reduced budgets.
- Update target setting algorithm in README for conformer.


sourabh2k15 (approved):
LGTM! thanks for catching this bug
hparam changes are also correct

PR #534, opened 2023-10-06
Adjust runtime budget for self-tuning ruleset and check that tuning search space is `None`

### LINKED ISSUE #531: Update submission runner to support self-tuning ruleset
In submission_runner.py if the tuning_ruleset flag is not external, check that tuning_search_space is None and extend the max runtime to 3x the budget for the external ruleset.

runame:
Resolves #531.


priyakasimbeg (approved):
Perfect thanks!

PR #535, opened 2023-10-06
Remove test target from scoring

### LINKED ISSUE #532: Update scoring code to ignore test targets
Given the update in the rules to exclude test targets from scoring we should update scoring.py. 
Specifically get_index_that_reaches_target should ignore the test target when it gets the index.

runame:
Resolves #532.

I simply removed everything related to the test set from scoring, hope this is what was intended.


priyakasimbeg (approved):
LGTM! Thanks!

Issue #512, opened 2023-09-15
Add support in scoring script for experiments ran w append_timestamps flag


priyakasimbeg:
Currently the scoring code takes in an experiment directory with the following structure:
+ experiment_directory_name
 + workload_name
  + trial_number
   - eval_measuremnts.csv

When the submissions are run with append_timestamp the experiment_directory_name will be experiment_directory_name_<timestamp>. Our scoring code currently does not support finding  and grouping eval_measurements.csv files in experiment directories with the same name but different timestamps. 

I think, ideally, we want to pass in an experiment_directory_name without timestamp. The scoring script should look for all experiment directories that match this name (with or without timestamp) and collect the eval_measurements.csv files as one experiment. 

Alternatively, we can instruct the submitter to organize files as the scoring script expects them.

priyakasimbeg:
Fixed in https://github.com/mlcommons/algorithmic-efficiency/pull/537

PR #537, opened 2023-10-09
Update scoring-utils to collect results from workdirs run with different timestamps

chandramouli-sastry:
Addresses #512. 
I simply considered directories sharing same prefix that are possibly different only in terms of the timestamp and updated the trial name column to now be a tuple of <trial, real_experiment_dir>. Does this solution seem okay? 

Although this addresses the problem, I was wondering if we should support append_timestamp as this also affects handling of preemption?


priyakasimbeg (changes requested):
@chandramouli-sastry thanks for working on this. This looks like a good solution to me. Before we check this off let's make sure its still compatible with the rest of the scoring API. Could you please add the following tests in this PR as well? 
1. In test_scoring_utils.py you'll find a test  called get_experiment_df that calls scoring_utils.get_experiment_df on TEST_DIR. Can you add a fake test_dir in test_data that has some dummy timestamp appended and a new test that takes in that test_dir with dummy timestamp. 
2. To check that the df is compatible w David's earlier scoring code, can we write a test that gets an experiment_df from this new fake test_dir w appended timestamp and call `scoring.compute_performance_profiles`. 

Also good question on compatibility with preemption. I think Zack and I talked about this when we added the append_timestamp flag and agreed at the time that we won't support preemption for experiment ran with append_timestamp, so no need to worry about that for now.

chandramouli-sastry:
Hi, thanks for your taking a loko! 

I added a test-directory that shares the same prefix and updated the tests. Is this what you asked for? 

Please let me know if I'm missing something!

priyakasimbeg (approved):

priyakasimbeg (dismissed):

priyakasimbeg:
@chandramouli-sastry there are some yapf and isort test failures after resolving merge conflicts w dev. I don't think I have permission to push to your repo branch. Could you resolve these formatting issues please?

PR #538, opened 2023-10-09
Update speech targets

priyakasimbeg:
Update speech targets after fixing mistake in internal procedure.


sourabh2k15 (changes requested):
can you verify internal deepspeech test/wer targets once ? everything else looks good

priyakasimbeg:
Checked that everything internally was correct.

sourabh2k15 (commented):
my bad test targets are also correct! approved

PR #544, opened 2023-10-11
Metadata logging fix

priyakasimbeg:
For some reason the meta data was not being written to a file.


algorithmic_efficiency/logger_utils.py
```diff
@@ -262,7 +262,7 @@ def _get_workload_properties(workload: spec.Workload) -> dict:
   return workload_properties
 
 
-def get_meta_data(workload: spec.Workload) -> dict:
+def get_meta_data(workload: spec.Workload, rng_seed: int = None) -> dict:
```
runame, line 265: I think using `dict` as a type hint is only supported for Python >= 3.9. So I think we can replace it with `Dict`, and when the types are known, we should also specify them, e.g. `Dict[str, int]`. Same thing for every use of `dict` as a type hint in this file.

algorithmic_efficiency/logger_utils.py
```diff
@@ -272,15 +272,11 @@ def get_meta_data(workload: spec.Workload) -> dict:
   meta_data.update(system_software_info)
   system_hardware_info = _get_system_hardware_info()
   meta_data.update(system_hardware_info)
+  if rng_seed:
```
runame, line 275: I think this should be `if rng_seed is not None:` because `rng_seed` could be `0` and then this condition what also evaluate to `False`.

algorithmic_efficiency/logger_utils.py
runame: Also, you can replace `rng_seed: int = None` with `rng_seed: Optional[int] = None`.

runame (approved):
Besides my comments, LGTM!

algorithmic_efficiency/logger_utils.py
priyakasimbeg: Thanks you're right. I replaced them with Dict for now.

algorithmic_efficiency/logger_utils.py
priyakasimbeg: That is a good point, will update this thanks!

PR #546, opened 2023-10-12
Suppress GPU not found warnings

### LINKED ISSUE #529: Misleading warnings on GPU not found
Looks like we often get GPU not found warnings, which are misleading in most cases since we hide the GPU. This issue has been filed to see if we can suppress the warnings to avoid confusing submitters.
Alternatively, we should document which warnings are expected in the README.

pomonam:
Fixes #529 (along with suppressing Tensorflow warnings).


priyakasimbeg (approved):
LGTM! Thanks you!

PR #558, opened 2023-10-30
Scoring fixes 

priyakasimbeg:
Scoring fixes, preliminary to issues/555, issues/556 and issues/557.

### DIFF
--- a/scoring/scoring.py
+++ b/scoring/scoring.py
@@ -129,7 +129,6 @@ def get_index_that_reaches_target(workload_df,
   op = operator.le if is_minimized else operator.ge
   validation_target_reached = validation_series.apply(
       lambda x: op(x, validation_target))
-  print(validation_target_reached)
   target_reached = pd.Series(validation_target_reached)
   # Remove trials that never reach the target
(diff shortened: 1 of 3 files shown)


scoring/scoring.py
```diff
@@ -34,6 +34,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+from tabulate import tabulate
```
runame, line 37: Where is this used?

runame (dismissed):
Besides my question, LGTM.

scoring/scoring.py
priyakasimbeg: Ah good catch, thanks, I used that for debugging I think. Removing.

setup.cfg
```diff
@@ -37,6 +37,7 @@ install_requires =
   absl-py==1.4.0
   numpy>=1.23
   pandas>=2.0.1
+  tabulate==0.9.0
```
runame, line 40: Is it used anywhere else in the repo? Couldn't find anything with a quick search.

runame (commented):
See my comment.

PR #559, opened 2023-10-31
Add flag for setting pytorch max_split_size_mb

priyakasimbeg:
The conformer pytorch workload still may run out of memory in current state. 
Add `reduce_pytorch_max_split_size` as a temporary workaround in case people into this issue. This will set 'max_split_size_mb:256'. Note that this will adversely impact the performance of the submission on this workload. See [tracking issue](https://github.com/mlcommons/algorithmic-efficiency/issues/497). 
Also minor formatting changes in other files.

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -150,7 +150,7 @@ flags.DEFINE_integer(
     'Value of rng seed. If None, a random seed will'
     'be generated from hardware.')
 flags.DEFINE_boolean('set_pytorch_max_split_size',
-                     None,
+                     False,
                      'If true, set pytorch max_split_size_mb to 256')
 FLAGS = flags.FLAGS
 USE_PYTORCH_DDP, RANK, DEVICE, N_GPUS = pytorch_setup()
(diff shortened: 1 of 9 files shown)


submission_runner.py
```diff
@@ -149,6 +149,9 @@
     None,
     'Value of rng seed. If None, a random seed will'
     'be generated from hardware.')
+flags.DEFINE_boolean('set_pytorch_max_split_size',
+                     None,
```
pomonam, line 153: Nit: We can set the default to `False`.

submission_runner.py
```diff
@@ -602,6 +605,9 @@ def main(_):
   if FLAGS.workload == 'librispeech_conformer':
     os.environ['XLA_PYTHON_CLIENT_MEM_FRACTION'] = '0.85'
 
+  if FLAGS.set_pytorch_max_split_size is True:
```
pomonam, line 608: Nit: Also, we can set `if FLAGS.set_pytorch_max_split_size: ...`.

pomonam (approved):
LGTM!

submission_runner.py
priyakasimbeg: Yup you're right that's better. Thanks!

submission_runner.py
priyakasimbeg: Fixed, thanks!

PR #561, opened 2023-11-03
Fix scoring issues

### LINKED ISSUE #555: Make scoring code consistent with scoring procedure
Currently the scoring code is not consistent with the scoring procedure. 
The scoring procedure for a submission:
1. For each workload look at 5 trials per workload.
2. If >=3 trials have reached the target, use the best trial out of the 5 trials in terms of submission time. 
3. If <3 trials have reached the target, the submission did not reach the target on that workload. 

In scoring.py we have to make the following modifications:
1. Print a warning if fewer than 5 trials are found for a workload submission.
2. Check that at 3/5 trials have reached the target and only then use the best trial to score the submission (bug fix)
3. Print a warning fewer than 8 workloads are found.
4. Make sure the percentage workloads that reached targets is calculated over all 8 workloads as opposed to the number of workloads in the submission directory. Currently if the user scores a submission over an experiment with n workloads it will use n as the numerator.

runame:
Resolves #555.

Please check if the fixes address all four points mentioned in the issue as intended.

**Notes:**
- I have removed the `tabulate` requirement, see my comment [here](https://github.com/mlcommons/algorithmic-efficiency/pull/558#discussion_r1380848399).
- I changed some import statements and file paths to absolute paths within the repo. Not sure if there was a reason to not do that previously, can undo if necessary.

### DIFF
--- a/scoring/score_submission.py
+++ b/scoring/score_submission.py
@@ -5,8 +5,7 @@ from absl import flags
 from absl import logging
 import scoring_utils
 
-from algorithmic_efficiency import workloads
-import scoring
+from scoring import scoring
 
 flags.DEFINE_string(
     'experiment_path',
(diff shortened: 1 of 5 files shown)


priyakasimbeg (approved):

priyakasimbeg:
LGTM! Thanks Runa!

PR #586, opened 2023-11-21
Prize qualification baselines

priyakasimbeg:
Add prize qualification baselines for external en self-tuning rulesets. 

@sourabh2k15 Could you double check that the parameter points are correct, they should correspond to internal opt list points 3, 4, 5, 6 and 8. 
Used nadamw with optlist point 8 for self-tuning ruleset.


reference_algorithms/threshold_baselines/README.md
```diff
@@ -0,0 +1,81 @@
+# Threshold Baselines
```
georgedahl, line 1: Let's call them "Prize qualification baselines"
Try to just spell out what they are in prose and only use that 3 word name in paths and folders and filenames.

georgedahl: We don't want people to confuse them with the qualification workloads/leaderboard, so it is good to also have the word "prize"

reference_algorithms/prize_qualification_baselines/external_tuning/tuning_search_space.json
```diff
@@ -0,0 +1,50 @@
+[
+    {
+        "dropout_rate": 0.0,
+        "label_smoothing": 0.1,
+        "learning_rate": 0.001308209823469072,
+        "one_minus_beta1": 0.02686663061,
+        "beta2": 0.9981232922116359,
```
sourabh2k15, line 7: beta1 values are missing, could you copy that over ?

sourabh2k15: nvm, I see one_minus_beta1 values

sourabh2k15 (approved):

reference_algorithms/threshold_baselines/README.md
priyakasimbeg: Fixed

PR #594, opened 2023-12-06
Add ImageNet ResNet Workload Variant

pomonam:
The PR adds 3 workload variants for ImageNet:
1. Change all ReLU activations to SiLU.
2. Change all ReLU activations to GELU.
3. Change the scale of the last BN initialization (for residual block) to 8. 

In addition, I added modeldiff tests for the first two variants, and they passed.


priyakasimbeg:
Thanks for getting this in! Just one request; could you add the variant workload names to this list in docker/scripts/startup.sh?

priyakasimbeg (approved):

PR #596, opened 2023-12-07
FastMRI workload variants

priyakasimbeg:
Add FastMRI workload variants:
- channels & pooling
- tanh
- layernorm

Also added modeldiff tests and checked that they passed.

Note on LayerNorm variant: We use torch.nn.GroupNorm with num_groups=1, to get an equivalent LayerNorm to JAX LayerNorm. The torch.nn.LayerNorm implementation only provides element-wise affine transformations, involving (C x W x H) size weights, whereas the JAX LayerNorm uses (C) size scaling weights. The torch.nn.GroupNorm module uses per-channel affine parameters which is equivalent to the Jax implementation.

### DIFF
--- a/algorithmic_efficiency/workloads/fastmri/fastmri_jax/models.py
+++ b/algorithmic_efficiency/workloads/fastmri/fastmri_jax/models.py
@@ -65,7 +65,7 @@ class UNet(nn.Module):
     if dropout_rate is None:
       dropout_rate = 0.0
 
-    # pylint: disable=C0103
+    # pylint: disable=invalid-name
     _ConvBlock = functools.partial(
         ConvBlock,
         dropout_rate=dropout_rate,
--- a/algorithmic_efficiency/workloads/fastmri/fastmri_jax/models.py
+++ b/algorithmic_efficiency/workloads/fastmri/fastmri_jax/models.py
@@ -28,7 +28,7 @@ def _instance_norm2d(x, axes, epsilon=1e-5):
   mean2 = jnp.mean(jnp.square(x), axes)
   # mean2 - _abs_sq(mean) is not guaranteed to be non-negative due
   # to floating point round-off errors.
-  var = jnp.maximum(0., mean2 - jnp.square(mean))  
+  var = jnp.maximum(0., mean2 - jnp.square(mean))
   stats_shape = list(x.shape)
   for axis in axes:
     stats_shape[axis] = 1
@@ -128,7 +128,7 @@ class UNet(nn.Module):
     output = nn.Conv(out_channels, kernel_size=(1, 1), strides=(1, 1))(output)
     return output.squeeze(-1)
 
-    
+
 class ConvBlock(nn.Module):
   """A Convolutional Block.
   out_channels: Number of channels in the output.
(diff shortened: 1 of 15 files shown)


algorithmic_efficiency/workloads/fastmri/fastmri_pytorch/workload.py
```diff
@@ -13,8 +13,7 @@
 from algorithmic_efficiency import pytorch_utils
 from algorithmic_efficiency import spec
 import algorithmic_efficiency.random_utils as prng
-from algorithmic_efficiency.workloads.fastmri.fastmri_pytorch.models import \
-    UNet
+from algorithmic_efficiency.workloads.fastmri.fastmri_pytorch import models
```
runame, line 16: Why not directly import `UNet`?

algorithmic_efficiency/workloads/fastmri/fastmri_jax/models.py
```diff
@@ -122,8 +125,8 @@ def __call__(self, x, train=True):
       output = jnp.concatenate((output, downsample_layer), axis=-1)
       output = conv(output, train)
 
-    output = final_conv(output)
-
+    out_channels = 1
```
runame, line 128: Why is this not a class attribute anymore?

algorithmic_efficiency/workloads/fastmri/fastmri_jax/workload.py
```diff
@@ -156,3 +162,32 @@ def _eval_model_on_split(self,
           k: v + synced_metrics[k][0] for k, v in total_metrics.items()
       }
     return {k: float(v.item() / num_examples) for k, v in total_metrics.items()}
+
+
+class FastMRIModelSizeWorkload(FastMRIWorkload):
```
runame, line 167: Are there not the adjusted validation and test targets missing here and for the Jax variants as well?

algorithmic_efficiency/workloads/fastmri/fastmri_pytorch/models.py
```diff
@@ -24,38 +43,49 @@ class UNet(nn.Module):
...
 
-    self.up_transpose_conv.append(TransposeConvBlock(ch * 2, ch))
+    self.up_transpose_conv.append(
+        TransposeConvBlock(ch * 2, ch, use_tanh, use_layer_norm))
     self.up_conv.append(
         nn.Sequential(
-            ConvBlock(ch * 2, ch, dropout_rate),
-            nn.Conv2d(ch, self.out_chans, kernel_size=1, stride=1),
+            ConvBlock(ch * 2, ch, dropout_rate, use_tanh, use_layer_norm),
+            nn.Conv2d(ch, 1, kernel_size=1, stride=1),
```
runame, line 88: Is there a reason why `out_chans` is hard-coded now?

runame (approved):
I haven't verified the equivalence of the different layer norm implementations, but besides my comments, LGTM!

algorithmic_efficiency/workloads/fastmri/fastmri_jax/workload.py
priyakasimbeg: Yes there are but we have to redo the target setting because of a bug we had in the init2winit code. I'll add this note in the PR.

algorithmic_efficiency/workloads/fastmri/fastmri_jax/models.py
priyakasimbeg: I forked this part from the old variants branch. I'm honestly not sure. Thanks for catching this, fixing.

algorithmic_efficiency/workloads/fastmri/fastmri_pytorch/models.py
priyakasimbeg: Nope. Fixed.

PR #599, opened 2023-12-09
Add workload variants for ImageNet ViT

pomonam:
This PR adds 3 variants for ImageNet ViT workloads:

1. Post-LN: Layer normalization applied after the residual branch. 
2. MAP: Applied multihead attention pooling. 
3. GLU: Included gated linear units in the MLPBlock.

In addition, I added modeldiff tests for the first two variants, and they passed (there was a bug in the test, and this was also fixed).


priyakasimbeg:
Hi Juhan, this looks great thank you! Were you planning on adding the modeldiff / fprop test for the MAP variant as well?

pomonam:
Thank you! Not yet - if you want to add the variants in the repo immediately, we can push this, and I will make a separate PR. But if it is not something urgent, I should be able to add modeldiff test next week.

priyakasimbeg (approved):

PR #606, opened 2024-01-05
Update reference_algorithms/prize_qualification_baselines/README.md

st--:
Fix a presumable typo - I'm assuming the self-tuning section ought to refer to the `self_tuning` directory.


priyakasimbeg:
Thanks for catching this. Will fix on dev.

PR #608, opened 2024-01-08
Comparator for VIT with attention pooling

chandramouli-sastry:
1. Includes a comparator for VIT with attention pooling and organizes the compare.py files of imagenet vit similar to other workloads.  
2. The bias for multihead attention was set to False in pytorch  and True in Jax: I changed the Jax code to make it identical to pytorch. @pomonam, did you intend to use biases for the pooling layer? After this change, both VIT implementations are identical.


priyakasimbeg (approved):
LGTM! Thanks for finishing this test and finding the source of the diff!

PR #613, opened 2024-01-13
Add traindiffs tests to workflows

### LINKED ISSUE #601: Include modeldiff tests in github workflows
## Description

We currently just print the result of the modeldiff tests. I think it would make sense to decide on thresholds for all workloads, such that we can make a binary pass/fail decision for each test. Then we should include the tests in our github workflows. This would guarantee that any future model changes won't lead to inconsistencies between Jax and PyTorch without the devs having to manually run anything.

runame:
Resolves #601.

Two things to pay attention to when reviewing:
1. The thresholds are set arbitrarily such that everything passes. They are pretty lenient, not sure if this will catch any future diffs.
2. I haven't tested the self-hosted workflow, but tried to follow the regression tests.


priyakasimbeg (dismissed):
LGTM! Thanks for adding the tests and making the style fixes in the imports! 
It looks like there are a couple of merge conflicts in the PR, could you resolve them?

priyakasimbeg (approved):

PR #625, opened 2024-02-02
Changes to random utils

priyakasimbeg:
- Update function signatures for _split and _fold_in to accurately reflect SeedType inputs and outputs.
- Change _fold_in to use data for sampling new seed instead of returning data.
- Force PyTorch function SeedType outputs to be 1-dim as opposed to 2-dim which is typical for Jax PRNGKeys
- Update run_workloads.py to use newly added randint for getting workload seeds.

### DIFF
--- a/algorithmic_efficiency/random_utils.py
+++ b/algorithmic_efficiency/random_utils.py
@@ -86,8 +86,8 @@ def PRNGKey(seed: SeedType) -> SeedType:  # pylint: disable=invalid-name
   return _PRNGKey(seed)
 
 
-def randint(seed:SeedType) -> int:
+def randint(seed: SeedType) -> int:
   if FLAGS.framework == 'jax':
     _check_jax_install()
-    return jax_rng.randint(seed, )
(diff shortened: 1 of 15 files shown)


algorithmic_efficiency/random_utils.py
```diff
@@ -58,7 +68,13 @@ def _check_jax_install() -> None:
         '--framework=pytorch to use the Numpy version instead.')
 
 
-def fold_in(seed: SeedType, data: Any) -> List[Union[SeedType, Any]]:
+def _bits(seed: SeedType) -> int:
+  rng = np.random.RandomState(_signed_to_unsigned(seed))
+  b = rng.bytes(4)
+  return int.from_bytes(b, byteorder='little')
```
georgedahl, line 74: Maybe just return the seed. It looks kind of silly, but given that bits is supposed to return an int you can "convert" information from an integer seed into an int trivially.

georgedahl (approved):

PR #640, opened 2024-02-15
Review scoring code & dataset setup

fsschneider:
I retroactively looked through PR #618 and also made some edits to the dataset setup readme. Specifically,

**Dataset**
- Changed all dataset sizes in the [dataset setup readme](https://github.com/mlcommons/algorithmic-efficiency/blob/main/datasets/README.md) to report apparent sizes via the `--apparent-size` option for `du`. This should make it more system-agnostic (e.g. ignore block sizes) when verifying whether the downloaded dataset is complete and consistent. @priyakasimbeg could you verify that the new sizes are consistent with your datasets?

**Rules**
- Clarified in the [scoring section](https://github.com/mlcommons/algorithmic-efficiency/blob/main/DOCUMENTATION.md#scoring) of the rules that we are only using the held-out workloads "if they exist" for this base fixed workload.

**Scoring code**
- Removed the `workload` flag in `generating_held_out_workloads.py` as it seemed irrelevant.
- It seems that the current `test_scoring_utils.py` is broken:
  - The imports `from scoring.scoring import NUM_TRIALS` and `from scoring.scoring import NUM_WORKLOADS` are failing for me. I think we should instead import `from scoring import performance_profile` and modify the rest of the code accordingly.
  - I had to modify the `scoring/test_data/` directory to get (some) tests to pass. E.g. I had to restructure them in a shared study folder.
  - Now 4/5 tests pass, but the performance profile fails, since there is no "base_workload" defined for the `mnist` workload (used in the `test_data/` (see https://github.com/mlcommons/algorithmic-efficiency/blob/11b11ec8160b0ca5b1503d85125420247fc58d16/algorithmic_efficiency/workloads/workloads.py#L134-L149).
  - Maybe, since we plan to upload the logs of the prize_qualification_baselines, we can just use those as our test data instead?

**Other**
- Other misc minor modifications, e.g. to the Getting Ready document.


priyakasimbeg:
I found 2 small discrepancies:
- imagenet/jax has 1113 files  vs 1111 files  on mine and your system respectively
- librispeech 388 GB vs 387 GB on mine and your system respectively (the number of file is the same).

At this stage we should just treat the output of the dataset up as the source of truth, so I will update the imagenet/jax and librispeech datasets on my VMs (unless I find some other bug in the process).

priyakasimbeg:
Regarding testing, thanks for looking at the test for the scoring utils. I will address remaining issues in a follow up PR for testing scoring code.

fsschneider:
> I found 2 small discrepancies:
> 
> * imagenet/jax has 1113 files  vs 1111 files  on mine and your system respectively
> * librispeech 388 GB vs 387 GB on mine and your system respectively (the number of file is the same).
> 
> At this stage we should just treat the output of the dataset up as the source of truth, so I will update the imagenet/jax and librispeech datasets on my VMs (unless I find some other bug in the process).

Regarding **librispeech**: I forgot to add the `--apparent-size` option for the `du` command in the readme for librispeech (but used it), perhaps this is the confusion. With this option, I get 387 GB and it also exactly matches what you reported for each individual folder:

Yours:
image

Mine:
image

Regarding **imagenet/jax**: I am not sure where the difference is coming from. For both the `imagenet2012` and the `imagenet_v2` folders there are always 3 files (`dataset_info.json`, `features.json`, `labels.labels.txt`) plus the tfrecord fils (64 validation, 1024 training, and 16 test). This is a total of 1110. Then I also have a `metadata.json` file in the `.config` folder in `imagenet_v2` making it a total of 1111 files.

priyakasimbeg (approved):

PR #656, opened 2024-02-27
Reduce batch size for imagenet variants

pomonam:


priyakasimbeg (commented):
I noticed the eval batch sizes are set to 1024. Just want to double check that these variant run fine with only the change in training bsz?

pomonam:
Yes, I checked that the eval batch size can stay the same.

Issue #684, opened 2024-03-05
Mismatch between target values in `workload.py` files and on `DOCUMENTATION.md` page


tfaod:
A few weeks ago, I noticed that the target values on the “getting started” are not consistent with the target values specified in the `workload.py` files. I thought this was a typo and previously I updated the `workload.py` files with the validation values. 

This week, I noticed that our training runs are hitting the target values specified in the workload.py but not the Getting Started documentation. For example, our fastmri workload consistently hits the test target, but never the val target from `DOCUMENTATION.md`. 

Could we confirm that the `workload.py` target files are correct, and fix this discrepancy? Thanks!!

priyakasimbeg:
Hi, we are working on updating the documentation with the correct targets and maximum runtimes.
In the meantime please refer to the workload.py files as the source of truth.

@runame

PR #685, opened 2024-03-05
Fix workload targets and max runtimes in docs

### LINKED ISSUE #660: Mismatch between Maximum Runtime in the DOCS and `max_allowed_runtime_sec` in the code
## Description

The "maximum runtime" of [this table](https://github.com/mlcommons/algorithmic-efficiency/blob/main/DOCUMENTATION.md#fixed-workloads) differ from `max_allowed_runtime_sec` in the code.

Few examples below:

Criteo:
https://github.com/mlcommons/algorithmic-efficiency/blob/a1bff1479fdbe7d8386c6a31bb75f42a19b0cacc/algorithmic_efficiency/workloads/criteo1tb/workload.py#L96

Deepspeech:
https://github.com/mlcommons/algorithmic-efficiency/blob/a1bff1479fdbe7d8386c6a31bb75f42a19b0cacc/algorithmic_efficiency/workloads/librispeech_deepspeech/librispeech_pytorch/workload.py#L84

ViT:
https://github.com/mlcommons/algorithmic-efficiency/blob/a1bff1479fdbe7d8386c6a31bb75f42a19b0cacc/algorithmic_efficiency/workloads/imagenet_vit/workload.py#L84

runame:
Fixes #660 and #684.

@priyakasimbeg I noticed that we don't have a shared parent class for the librispeech deepspeech workloads and specify the targets, max runtime, and other properties twice -- once for Jax and once for PyTorch. Is there any reason for that? If not we should probably create a separate workload parent class. Otherwise it is easy to introduce a bug by accidentally making these values inconsistent.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -400,12 +400,12 @@ The currently eight fixed workloads are:
 
 |            | **Task**                      | **Dataset** | **Model**               | **Loss** | **Metric** | Validation<br>**Target** | Test<br>**Target**   | Maximum<br>**Runtime** <br>(in secs) |
 |------------|-------------------------------|-------------|-------------------------|----------|------------|--------------------------|----------------------|------------------------|
-| **1**      | Clickthrough rate prediction  | Criteo 1TB  | DLRMsmall               | CE       | CE         | 0.123649                 | 0.126060                |       21,600                 |
-| **2**      | MRI reconstruction            | fastMRI     | U-Net                   | L1       | SSIM       | 0.7344                   | 0.741652             |          10,800              |
-| **3<br>4** | Image classification          | ImageNet    | ResNet-50<br>ViT        | CE       | ER         | 0.22569<br>0.22691        | 0.3440<br>0.3481    |        111,600    <br> 111,600            |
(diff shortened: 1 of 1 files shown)


priyakasimbeg:
Hi Runa,
Thanks a ton for working on this.

> I noticed that we don't have a shared parent class for the Librispeech DeepSpeech workloads and specify the targets, max runtime, and other properties twice -- once for Jax and once for PyTorch. Is there any reason for that? 

Right they inherit from the conformer JAX and PyTorch workloads to reuse the input pipeline  We actually had a bug until a few months ago where the DeepSpeech workloads were inheriting all the properties from the conformer workloads (including targets). The targets and runtimes were specified in a Librispeech DeepSpeech parent workload that was not being used so I removed it (https://github.com/mlcommons/algorithmic-efficiency/pull/526).  
At the time I took a stab at separating DeepSpeech from Conformer but it quickly turned into a mess with a large amount of duplicate code. 
This is definitely a good thing to fix at some point in the future but I don't think we want to go down that rabbit hole right now.

priyakasimbeg (approved):

Issue #638, opened 2024-02-14
num_workers=0 in ImageNet DataLoader


Niccolo-Ajroldi:
This is more of a question than an issue.

The **number of workers** for torch `DataLoader` is set to zero for ImageNet during evaluation (on `train`, `valid` and `test` splits). However, this makes evaluation extremely slow in my setting.

Maybe I am missing something, hence I would like to ask:
- Why is the number of workers set to zero during evaluation?
- Are we allowed to change the number of workers for evaluation to be 4 when self-reporting results? 
Under the rules, model evaluations should be untimed and not contribute to the submissions score, so this shouldn't influence the final score. Is this correct?

Here is the relevant code:
https://github.com/mlcommons/algorithmic-efficiency/blob/11b11ec8160b0ca5b1503d85125420247fc58d16/algorithmic_efficiency/workloads/imagenet_resnet/imagenet_pytorch/workload.py#L129

Notice that for other datasets like _librispeech_, the number of workers is always 4:
https://github.com/mlcommons/algorithmic-efficiency/blob/11b11ec8160b0ca5b1503d85125420247fc58d16/algorithmic_efficiency/workloads/librispeech_conformer/librispeech_pytorch/workload.py#L176

Why this discrepancy?

Thanks in advance!

priyakasimbeg:
@pomonam do you have any context around this?

pomonam:
Sorry for the late reply! Yes, changing `num_workers` should not affect the final results because the evaluation time is not measured. Please feel free to change these numbers when running your experiments.

priyakasimbeg:
@runame added a flag to control the num_workers for some datasets during eval. 
Hope that helps!

Niccolo-Ajroldi:
Thank you, @runame!

PR #687, opened 2024-03-06
Add `pytorch_num_workers` flag for workloads with PyTorch data loaders

runame:
Address #638.

I have added a new flag to set `num_workers` for workloads with PyTorch data loaders. It defaults to 4 (same as before, with the exception of ImageNet evaluation). I think this flag makes sense because there is no general rule to set `num_workers` and users might want to tune it to improve performance on their hardware. Actually, the optimal number is probably different for each workload, but I think allowing for a per-workload setting adds too much complexity -- we can revisit this if users request it.

### DIFF
--- a/algorithmic_efficiency/data_utils.py
+++ b/algorithmic_efficiency/data_utils.py
@@ -231,7 +231,7 @@ def cycle(iterable: Iterable,
 class PrefetchedWrapper:
 
   def __init__(self,
-               dataloader: torch.utils.data.DataLoader,
+               dataloader: DataLoader,
                device: torch.device,
                start_epoch: int = 0) -> None:
     self.dataloader = dataloader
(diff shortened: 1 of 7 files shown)


priyakasimbeg (changes requested):
Thanks for working on this!

Could we modify this flag to only apply to eval mode? 
I think we want to restrict any changes through submission_runner that may affect the data loading speed in train steps because  data_selection is included in the submission time.

runame:
Just changed the flag to only apply to eval. I thought it might make sense if people can tune it for training while developing their submission, but you are right that this might lead to errors when self-reporting results.

priyakasimbeg (approved):
LGTM! Thanks Runa, this is a very nicely thought out and user-friendly solution :).

PR #696, opened 2024-03-10
Fix check for `eval_num_workers` property

runame:


priyakasimbeg (approved):
LGTM! Thanks for the quick fix!

PR #707, opened 2024-03-16
Fix typo in `jax_nadamw_target_setting.py`

tfaod:
Fix dict indexing typo in `jax_nadamw_target_setting.py`


priyakasimbeg:
Thanks for catching this!

Issue #663, opened 2024-02-29
LibriSpeech Conformer Workload OOMs or NCCL Errors When Run With Multiple Trials


hjmshi:
We consistently observe an OOM error when running the one of the NAdamW baselines on LibriSpeech Conformer with multiple trials in PyTorch. This is run for the external ruleset. The first trial will successfully run through, but any subsequent trial will OOM.

If we try to resume a multi-trial run, we will observe a NCCL error. This occurs even if we delete the `trial_2` folder (but the `trial_1` folder remains intact).  

## Description

As discussed above, we will observe OOM when running LibriSpeech Conformer with the NAdamW baseline with multiple trials on 8 V100s with 16GB each. This is an example of an OOM we observe on the subsequent trial:
```
I0229 11:25:52.179392 139670264227648 submission_runner.py:314] Starting training loop.                                                [1007/1887]
I0229 11:25:54.777722 139642405861120 logging_writer.py:48] [0] global_step=0, grad_norm=56.284924, loss=31.354658                                
I0229 11:25:54.783623 139670264227648 pytorch_nadamw_full_budget.py:296] 0) loss = 31.355, grad_norm = 56.285                                     
I0229 11:25:55.234757 139670264227648 spec.py:321] Evaluating on the training split.                                                              
I0229 11:26:08.470401 139670264227648 spec.py:333] Evaluating on the validation split.                                                            
I0229 11:26:20.220777 139670264227648 spec.py:349] Evaluating on the test split.                                                                  
I0229 11:26:26.181465 139670264227648 submission_runner.py:414] Time since start: 34.00s,       Step: 1,        {'train/ctc_loss': 30.998378480849
826, 'train/wer': 1.253173088944333, 'validation/ctc_loss': 29.436887863844042, 'validation/wer': 1.157427702409115, 'validation/num_examples': 53
48, 'test/ctc_loss': 29.538330745441247, 'test/wer': 1.1866837283935572, 'test/num_examples': 2472, 'score': 2.605729579925537, 'total_duration': 
34.00228404998779, 'accumulated_submission_time': 2.605729579925537, 'accumulated_eval_time': 30.946528434753418, 'accumulated_logging_time': 0}  
I0229 11:26:26.201718 139642405861120 logging_writer.py:48] [1] accumulated_eval_time=30.946528, accumulated_logging_time=0, accumulated_submissio
[...]

PR #710, opened 2024-03-19
Aiming to Fix Conformer OOM

pomonam:
Tries to address #663.
- Resets the memory at the beginning of training.
- Change dropout to have `inplace=True`


priyakasimbeg:
Hey Juhan, can you address the lint errors so we can merge this in?

PR #713, opened 2024-03-21
Don't save final checkpoint when `save_checkpoints=False`

### LINKED ISSUE #705: Add flag to completely opt out of checkpointing
## Description
Currently --save_checkpoints flag is only used to control saving intermediate checkpoints. 
Our checkpointing code currently is not compatible with all possible submissions, so we will need a flag to completely disable checkpointing. 

We should also clarify in the technical documentation that during scoring we will not use checkpoints.

runame:
Fixes #705.

I noticed that we already have a `save_intermediate_checkpoints` flag. When setting this to `False` and `save_checkpoints=True`, we can recreate the previous behaviour of only storing one final checkpoint.

Not sure where exactly in the docs to mention the lack of checkpointing during scoring? It shouldn't matter to the submitter as it is not timed, besides in edge cases like having an optimizer that does something like [this](https://github.com/facebookresearch/optimizers/blob/main/distributed_shampoo/distributed_shampoo.py#L901-L909), which would lead to an error in the submission if `save_checkpoints=True`.

### DIFF
--- a/submission_runner.py
+++ b/submission_runner.py
@@ -111,8 +111,8 @@ flags.DEFINE_string('experiment_name', None, 'Name of the experiment.')
 flags.DEFINE_boolean(
     'save_checkpoints',
     True,
-    'Whether or not to save checkpoints of the model at every eval and after training.',
-)
+    'Whether or not to save checkpoints of the model and optimizer '
+    'at every eval and after training.')
 flags.DEFINE_boolean(
     'save_intermediate_checkpoints',
     True,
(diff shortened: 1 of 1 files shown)


priyakasimbeg:
> I noticed that we already have a save_intermediate_checkpoints flag.
I forgot about that. 

> Not sure where exactly in the docs to mention the lack of checkpointing during scoring?  
I'm not sure what the best place is either. I think we can add a question to the FAQS in DOCUMENTATION.md like: "my optimizer is incompatible with the AlgoPerf checkpointing code, will this affect my submission?"

priyakasimbeg (approved):
I'm going to merge this into dev and we can add the documentation clarification in a separate PR.

PR #739, opened 2024-03-28
[fix] random_utils.py to `_signed_to_unsigned`

tfaod:
When running the submission_runner on the self-tuning track, we run into this error calling `_signed_to_unsigned` from random_utils.py. 

I've added a fix

```ValueError: Seed must be between 0 and 2**32 - 1
    rng = prng.PRNGKey(rng_seed)
  File "/private/home/axyang/optimization/algorithmic-efficiency-entry/algorithm
ic_efficiency/random_utils.py", line 79, in PRNGKey
    return _PRNGKey(seed)
```


tfaod:
Error comes from the following line in the self-tuning infra:
https://github.com/mlcommons/algorithmic-efficiency/blob/576d5e37c318f1d04398206ed2a781dd1dac0a56/submission_runner.py#L558

priyakasimbeg (approved):
LGTM thanks for the fix!

PR #755, opened 2024-04-06
Allow DDP checkpointing

pomonam:
This PR also suppresses the NCCL error message when the user types "N" to resume training.


priyakasimbeg (approved):
LGTM! Thanks for fixing this!

Issue #758, opened 2024-04-11
Support for optimizers that require an `.eval()` step 


netw0rkf10w:
## Description

First of all if this feature is already supported then please consider this as a question.

I'm trying to reproduce some results of existing algorithms such as SGD, AdamW, and also the recently proposed ScheduleFree. For the latter, in particular, there is an `.eval()` step that needs to be done before the validation phase, and I haven't figured out how to do that properly (there doesn't seem to be any indication in their [submission code](https://github.com/facebookresearch/schedule_free/blob/main/schedulefree/algoperf/external_tuning/schedule_free_adamw/submission.py)).

Any help would be greatly appreciated! Thank you in advance!

adefazio:
I use the closure form of the algorithm to avoid needing to do the eval() call. There has been a request from another user to support this this sort of eval() mode so that exponential weight averaging can be implemented. The organizers said that this would be something they will look at adding in the future. It should help with performance.

netw0rkf10w:
Thanks a lot, @adefazio. Nice trick! 

As I understand, you basically swap the extrapolated points at every step, is that correct? That seems to be putting your algorithm at a disadvantage though. Have you observed considerable slowdown compared to the `.eval()` version?

And I'm quite surprised that the proposed feature hasn't been added to support your method. It's just two lines of code (actually we only need **one**). Or maybe I'm missing something? 

Would love to hear your opinion on this, @priyakasimbeg. Thanks.

adefazio:
The overhead is minimal, parameter copying is very fast. 
The issue was raised very close to the deadline and so given the low overhead of the copy, it was decided that changing the competition API close to the deadline was not warranted.

priyakasimbeg:
Hi we're planning on discussing this feature request on Th 9/5 during the WG meeting

priyakasimbeg:
I believe this is partially addressed in https://github.com/mlcommons/algorithmic-efficiency/pull/798 (decouple BN statistic updates from using running statistics) and will be fully addressed once pr/789 for https://github.com/mlcommons/algorithmic-efficiency/issues/719 has been submitted. So I am deduping this. Please feel free to reopen if I have missed anything.

Issue #785, opened 2024-09-12
Inform submission about `accumulated_submission_time`


Niccolo-Ajroldi:
## Description

Currently, `update_params` has no up-to-date information about the elapsed time since start. 

My motivation for adding this feature is to simplify the implementation of a _time-based learning rate schedule_.

**Can't a submission just keep track of time or estimate it?**
In theory yes, this is allowed by the rules and feasible. However, such implementation would require synchronization among devices inside `update_params` when training in distributed mode, which would penalize such submission.

**Why is a time-based scheduler useful?**
Currently, a submission can implement a LR scheduler using `step_hint` as a step budget. This is a reliable estimate of the number of steps needed for (N)AdamW to reach max_runtime. However, a submission could be faster/slower than (N)AdamW, and the extent of this difference can vary based on the workload itself. This makes deriving a custom step budget from `step_hint` suboptimal.

## Implementation
We could simply pass `train_state` to `update_params`, or even just `train_state['accumulated_submission_time']`.
Notice that `update_params` already receives in input `eval_results`, which informs the submission about the elapsed time at the moment of the last evaluation, which is not up-to date with `accumulated_submission_time`.

Niccolo-Ajroldi:
Remark: we should avoid that a submission modifies 'train_state'.
A straightforward solution would be to pass it by copy.

priyakasimbeg:
Fixed in https://github.com/mlcommons/algorithmic-efficiency/pull/790

PR #789, opened 2024-09-15
Introduce prepare for eval, fix evaluation bug

### LINKED ISSUE #719: Inform submission about evaluation step
**tl;dr**: We should let the submission know if an evaluation is going to happen at the current step or not.

## Description
Currently, there is no easy way for the submission to know if the model returned by `update_params`
is going to be evaluated on the train/valid/test set. 

This limits the space of possible submission, or at least force them to apply some workarounds to infer whether the current step is going to be an evaluation step or not.  (A possible workaround is to keep track of time from last evaluation _inside the submission_, but this adds a non-negligible overhead to the submission itself and deviates from the original goal of the submission.).

An example of a submission where it is crucial to know when evaluation is going to occur is **Stochastic Weight Average**.

## Possible solutions
A straightforward solution is to decide if the submission is eligible for an untimed eval **before** calling `update_params`, and add an argument to `update_params` that passes this information to the submission.

The only drawback with this approach is that we don't evaluate every `workload.eval_period_time_sec`, but a little less frequently (we evaluate every `workload.eval_period_time_sec` + _submission_time_per_step_). Assuming that `workload.eval_period_time_sec` >> _submission_time_per_step_, this is hopefully not a big difference.

I think this is an important feature, and it would be nice to implement it.

Niccolo-Ajroldi:
## Description
This pull request introduces a `prepare_for_eval` function and updates the code to support it.

The implementation follows the blueprint of @fsschneider in https://github.com/mlcommons/algorithmic-efficiency/issues/719#issuecomment-2328797610 and fixes the bug of giving a free evaluation to a submission that goes out of max_runtime (again https://github.com/mlcommons/algorithmic-efficiency/issues/719#issuecomment-2328797610).

## Function signature

The arguments of `prepare_for_eval` are the same as `update_params`, except for `batch`. I believe that `prepare_for_eval` should indeed be agnostic to the last batch used during training. The return type is the same as `update_params`.

```python
def prepare_for_eval(workload: spec.Workload,
                     current_param_container: spec.ParameterContainer,
                     current_params_types: spec.ParameterTypeTree,
                     model_state: spec.ModelAuxiliaryState,
                     hyperparameters: spec.Hyperparameters,
                     loss_type: spec.LossType,
                     optimizer_state: spec.OptimizerState,
                     eval_results: List[Tuple[int, float]],
                     global_step: int,
                     rng: spec.RandomState) -> spec.UpdateReturn:
  return (optimizer_state, current_param_container, model_state)
```

## List of changes
In `submission_runner.py`:
- add timed call to `prepare_for_eval`
- add `profiler`
- move `del batch` before `prepare_for_eval` (instead than before evaluation)
- update `accumulated_submission_time` after `prepare_for_eval`
- compute `is_time_remaining` after `prepare_for_eval`
- proceed to eval iff `is_time_remaining`
- add `prep_eval_rng`

Minor changes:
- add `PrepareForEvalFn` to `spec`
- add `prepare_for_eval` to submission template
- add `prepare_for_eval` to all pytorch and jax submissions
- update the docs

Fixes #719  and #758 .


submission_runner.py
```diff
@@ -518,6 +548,7 @@ def score_submission_on_workload(workload: spec.Workload,
   init_optimizer_state = submission_module.init_optimizer_state
   update_params = submission_module.update_params
   data_selection = submission_module.data_selection
+  prepare_for_eval = submission_module.prepare_for_eval
```
priyakasimbeg, line 551: This will break the submission modules of the competition. Should we default this to a no op function maybe in a try except block?

Niccolo-Ajroldi: It depends if we want to make the code backward compatible with the previous submissions. If so, we could make 
`prepare_for_eval` a no-op by doing smth like this:
```python
prepare_for_eval = getattr(submission_module, 'prepare_for_eval', None)
```

Ant then call it only if it is not `None`:
```python
# Prepare for evaluation (timed).
if prepare_for_eval is not None:

  with profiler.profile('Prepare for eval'):
    del batch
    prepare_for_eval_start_time = get_time()
    optimizer_state, model_params, model_state = prepare_for_eval(...)
    prepare_for_eval_end_time = get_time()

  # Update sumbission time.
  train_state['accumulated_submission_time'] += (
      prepare_for_eval_end_time - prepare_for_eval_start_time)
```
What do you think?


My only doubt is that I am not sure that we can make all the API changes backward compatible to the submissions, for example #785 is more tricky to implement in such a way.

priyakasimbeg: At the moment, I think backwards compatibility is important since we are transitioning to the rolling leaderboard (which may involve hardware changes) and will have to potentially rescore some submissions. 
I don't think we want to bear the responsibility to modify the existing submissions' code.
So I prefer this solution to get the prepare_for_eval function and call it only it if it is not None.

Niccolo-Ajroldi: I agree, it makes sense! I changed the code in this regard and tested it for backward compatibility, it looks good.

priyakasimbeg:
@fsschneider can we get your LGTM on this?

fsschneider (approved):
Sorry for the very late review.
It looks good to me! Thanks for all the work Niccolò!

Quick questions, the changes to the `reference_algorithms` and `prize_qualification` algorithms aren't strictly necessary, right? Since we now default to a no-op if `prepare_for_eval` is not defined.
I fully agree that making it backward-compatible in that way is really helpful.

Again thanks for this crucial PR!

Niccolo-Ajroldi:
Thank you for reviewing!

Correct, the updates to `reference_algorithms` and `prize_qualification` are not necessary thanks to backward-compatibility. 

I figured it would be nice to have all the implemented submissions reflecting the latest stage of the code, but I can reverse those changes if needed! (:

fsschneider:
> I figured it would be nice to have all the implemented submissions reflecting the latest stage of the code, but I can reverse those changes if needed! (:

I agree with you. No need to revert, just wanted to check if I understand it correctly :)

PR #790, opened 2024-10-03
Pass `train_state `to `update_params`

Niccolo-Ajroldi:
This pull-request adds `train_state` to the arguments of `update_params`. 

This allows for example to implement a time-based LR scheduler (see #785).

Since `train_state` is a dictionary, we have to avoid that a submission modifies its content.
The solution adopted here passes `train_state` by copy. We use a shallow copying, since all values in `train_state` are primitive types, and cannot be modified in-place. Although very fast, this copying operation would however be counted as part of the submission time.

A possible alternative solution that would avoid this problem would be to pass `train_state['accumulated_submission_time']` directly instead of passing the entire dictionary.

### DIFF
--- a/algorithmic_efficiency/spec.py
+++ b/algorithmic_efficiency/spec.py
@@ -401,10 +401,10 @@ UpdateParamsFn = Callable[[
     Dict[str, Tensor],
     LossType,
     OptimizerState,
-    Dict[str, Any],
     List[Tuple[int, float]],
     int,
-    RandomState
+    RandomState,
+    Optional[Dict[str, Any]]
 ],
                           UpdateReturn]
 
@@ -423,10 +423,10 @@ def update_params(workload: Workload,
                   batch: Dict[str, Tensor],
                   loss_type: LossType,
                   optimizer_state: OptimizerState,
-                  train_state: Dict[str, Any],
                   eval_results: List[Tuple[int, float]],
                   global_step: int,
-                  rng: RandomState) -> UpdateReturn:
+                  rng: RandomState,
+                  train_state: Optional[Dict[str, Any]] = None) -> UpdateReturn:
   """Return (updated_optimizer_state, updated_params, updated_model_state)."""
   pass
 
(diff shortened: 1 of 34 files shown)


algorithmic_efficiency/spec.py
```diff
@@ -422,6 +423,7 @@ def update_params(workload: Workload,
                   batch: Dict[str, Tensor],
                   loss_type: LossType,
                   optimizer_state: OptimizerState,
+                  train_state: Dict[str, Any],
```
priyakasimbeg, line 426: Can we make this an optional kwarg and default it to None for backwards compatibility?

priyakasimbeg: train_state is just a small dict[str, float] right? does it take a significant amount of time to copy it?

Also, good point about checking before passing the arg to the train_state. I guess the code may look a little weird to people who don't have context about the backward compatiliblity. But I can't think of a cleaner solution that also maintains the backwards compatibility, so let's add your suggestion.
Alternatively we could make a small change to all submissions and add **kwargs to update_params. But I feel like that is a slippery slope. WDYT?

Niccolo-Ajroldi: Yes, `train_state` is very small, it should add a minimal overhead indeed, I just didn't like the idea of having extra ops in the submission time 😅. But we could also pass a `MappingProxyType(train_state)` to pass a read-only version without need for copying, which should be faster.

Regarding the last point, I agree with you. We can start like this and perhaps if we later want to simplify the logic at the cost of adding **kwargs, we can do that.

priyakasimbeg: Ok, let's go ahead with the `inspect` route and I will merge this PR in

Niccolo-Ajroldi: Ok, I changed it, and it should be good to go now!

priyakasimbeg (approved):
lgtm

PR #792, opened 2024-10-11
Fix scoring bug in `get_summary_df`

### LINKED ISSUE #791: Scoring logging bug 🐛 -  incorrect computation of `time to target on val (s)` in `get_summary_df`
## Description

In `scoring/score_submissions.py`, the function `get_summary_df` is responsible for gathering evaluation statistics from submission logs into a DataFrame. When scoring a submission, this function is invoked on every workload, and the resulting concatenated DataFrames are saved as CSV files (`<submission>_summary.csv` ).

The current implementation computes the time needed to reach the validation target as follows: 
https://github.com/mlcommons/algorithmic-efficiency/blob/a23b5ea298acdfae7366fb6ca175ab68f3a38f91/scoring/score_submissions.py#L91-L94

This results in a `time to target on val (s)` equal to `time to best eval on val (s)` if a submission reaches the target.  However, usually the time to the validation target is usually lower than the time to best eval score.

## Performance profiles are not affected
Fortunately, this bug **does not affect** the performance profiles, nor the final scores. Despite the concatenated DataFrames are used to compute the performance profiles, fortunately, we ignore the existing `time to target on val (s)` column and perform instead a correct computation of the time to eval target.

## Source or Possible Fix

I have implemented a fix in #792. The final scores and the performance profiles are unaffected after the fix. However, `<submission>_summary.csv` changes drastically. Here is an example on two workloads for the prize qualification baseline algorithm (first study):

image

Niccolo-Ajroldi:
Fixes #791, correcting the computation of `time to target on val (s)` when scoring submissions.


Niccolo-Ajroldi:
Ah ops, yeah, it was left from some test that I ran, now it should be gone!

priyakasimbeg (approved):

Issue #805, opened 2024-10-31
Upgrade to Python 3.10


priyakasimbeg:
## Description

Upgrade our package for compatibility with python 3.10.
This will involve:

- Upgrading all the package dependencies to the latest compatible versions with python 3.10
- Updating our Docker images
- Updating the documentation to reflect that we are using Python 3.10

PR #810, opened 2024-11-14
fix: changing the dtype in random_utils to uint32

init-22:
#801

Issue #801, opened 2024-10-26
TypeError: JAX encountered invalid PRNG key data: expected key_data.dtype = uint32; got dtype=int32


init-22:
I was trying to run the submission_runner.py file inside the docker and got a TypeError,
Use these commands to reproduce the error:
```
sudo docker run -it   -v <PATH>/algorithmic-efficiency:/algorithmic-efficiency --runtime=nvidia algoperf_pytorch /bin/bash

cd algorithmic-efficiency

python3 submission_runner.py \
    --framework=pytorch \
    --workload=mnist \
    --experiment_dir=$HOME/experiments \
    --experiment_name=my_first_experiment \
    --submission_path=reference_algorithms/paper_baselines/adamw/jax/submission.py \
    --tuning_search_space=reference_algorithms/paper_baselines/adamw/tuning_search_space.json
```

Here is the traceback:

```
Traceback (most recent call last):
  File "submission_runner.py", line 714, in <module>
    app.run(main)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "/usr/local/lib/python3.8/dist-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
  File "submission_runner.py", line 682, in main
    score = score_submission_on_workload(
  File "submission_runner.py", line 587, in score_submission_on_workload
    timing, metrics = train_once(workload, workload_name,
  File "submission_runner.py", line 351, in train_once
    optimizer_state, model_params, model_state = update_params(
  File "/algorithmic-efficiency/reference_algorithms/paper_baselines/adamw/jax/submission.py", line 130, in update_params
    per_device_rngs = jax.random.split(rng, jax.local_device_count())
  File "/usr/local/lib/python3.8/dist-packages/jax/_src/random.py", line 217, in split
    key, wrapped = _check_prng_key(key)
  File "/usr/local/lib/python3.8/dist-packages/jax/_src/random.py", line 79, in _check_prng_key
    return prng.random_wrap(key, impl=default_prng_impl()), True
  File "/usr/local/lib/python3.8/dist-packages/jax/_src/prng.py", line 907, in random_wrap
    _check_prng_key_data(impl, base_arr)
  File "/usr/local/lib/python3.8/dist-packages/jax/_src/prng.py", line 119, in _check_prng_key_data
    raise TypeError("JAX encountered invalid PRNG key data: expected key_data.dtype = uint32; "
[...]

PR #811, opened 2024-11-15
Moving from Python3.8 to Python 3.11

init-22:
Making the following major changes:

* Python 3.11.10
* Jax 0.4.35
* Pytorch 2.5.1
* Tensorflow 2.18

#805

### DIFF
--- a/algorithmic_efficiency/random_utils.py
+++ b/algorithmic_efficiency/random_utils.py
@@ -18,8 +18,8 @@ FLAGS = flags.FLAGS
 
 # Annoyingly, RandomState(seed) requires seed to be in [0, 2 ** 32 - 1] (an
 # unsigned int), while RandomState.randint only accepts and returns signed ints.
-MAX_INT32 = 2**31
-MIN_INT32 = -MAX_INT32
+MAX_UINT32 = 2**31
+MIN_UINT32 = 0
 
 SeedType = Union[int, list, np.ndarray]
 
@@ -35,13 +35,13 @@ def _signed_to_unsigned(seed: SeedType) -> SeedType:
 
 def _fold_in(seed: SeedType, data: Any) -> List[Union[SeedType, Any]]:
   rng = np.random.RandomState(seed=_signed_to_unsigned(seed))
-  new_seed = rng.randint(MIN_INT32, MAX_INT32, dtype=np.int32)
+  new_seed = rng.randint(MIN_UINT32, MAX_UINT32, dtype=np.uint32)
   return [new_seed, data]
 
 
 def _split(seed: SeedType, num: int = 2) -> SeedType:
   rng = np.random.RandomState(seed=_signed_to_unsigned(seed))
-  return rng.randint(MIN_INT32, MAX_INT32, dtype=np.int32, size=[num, 2])
+  return rng.randint(MIN_UINT32, MAX_UINT32, dtype=np.uint32, size=[num, 2])
 
 
 def _PRNGKey(seed: SeedType) -> SeedType:  # pylint: disable=invalid-name
--- a/algorithmic_efficiency/random_utils.py
+++ b/algorithmic_efficiency/random_utils.py
@@ -18,7 +18,7 @@ FLAGS = flags.FLAGS
 
 # Annoyingly, RandomState(seed) requires seed to be in [0, 2 ** 32 - 1] (an
 # unsigned int), while RandomState.randint only accepts and returns signed ints.
-MAX_UINT32 = 2**31
+MAX_UINT32 = 2**32-1
 MIN_UINT32 = 0
 
 SeedType = Union[int, list, np.ndarray]
@@ -26,11 +26,11 @@ SeedType = Union[int, list, np.ndarray]
 
 def _signed_to_unsigned(seed: SeedType) -> SeedType:
   if isinstance(seed, int):
-    return seed % 2**32
+    return seed % MAX_UINT32
   if isinstance(seed, list):
-    return [s % 2**32 for s in seed]
+    return [s % MAX_UINT32 for s in seed]
   if isinstance(seed, np.ndarray):
-    return np.array([s % 2**32 for s in seed.tolist()])
+    return np.array([s % MAX_UINT32 for s in seed.tolist()])
 
 
 def _fold_in(seed: SeedType, data: Any) -> List[Union[SeedType, Any]]:
@@ -75,5 +75,5 @@ def split(seed: SeedType, num: int = 2) -> SeedType:
 def PRNGKey(seed: SeedType) -> SeedType:  # pylint: disable=invalid-name
   if FLAGS.framework == 'jax':
     _check_jax_install()
-    return jax_rng.PRNGKey(seed)
+    return jax_rng.key(seed)
   return _PRNGKey(seed)
--- a/algorithmic_efficiency/random_utils.py
+++ b/algorithmic_efficiency/random_utils.py
@@ -18,7 +18,7 @@ FLAGS = flags.FLAGS
 
 # Annoyingly, RandomState(seed) requires seed to be in [0, 2 ** 32 - 1] (an
 # unsigned int), while RandomState.randint only accepts and returns signed ints.
-MAX_UINT32 = 2**32-1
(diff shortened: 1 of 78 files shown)


priyakasimbeg:
temporarily changed target branch to main to trigger regression tests.

init-22:
Some of the required package versions are only available in python3.11 and the tests are using python3.9 it seems, we might need to change that

priyakasimbeg:
Could you update the tests to use 3.11 in this PR?

priyakasimbeg:
Hey Isaac, 
Could you paste the output of pip freeze in this PR as well  of an environment setup with these upgrades?

priyakasimbeg:
Running  Containerized regression tests: https://github.com/mlcommons/algorithmic-efficiency/actions/runs/11934647960

priyakasimbeg:
@init-22 I'm going to merge this in to a new branch mlcommons/python_upgrades so that I can trigger the tests. Currently the containerized tests on our self-hosted runners can only run branches that are on this repo. Could you pull that branch and make follow up changes on there?

algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py
```diff
@@ -12,10 +12,11 @@
 import jax.numpy as jnp
 import numpy as np
 import optax
+import sacrebleu
```
runame, line 15: What was the reason for switching to `sacrebleu` again?

runame (commented):
What was the reason for switching to `sacrebleu`?

I haven't reviewed the Dockerfile and all the version updates, the rest LGTM.

algorithmic_efficiency/random_utils.py
```diff
@@ -18,30 +18,30 @@
 
 # Annoyingly, RandomState(seed) requires seed to be in [0, 2 ** 32 - 1] (an
 # unsigned int), while RandomState.randint only accepts and returns signed ints.
-MAX_INT32 = 2**31
-MIN_INT32 = -MAX_INT32
+MAX_UINT32 = 2**32 - 1
```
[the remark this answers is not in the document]
priyakasimbeg: I think we started getting typeErrors for some reason when running after fresh installs: https://github.com/mlcommons/algorithmic-efficiency/issues/801. 

The same fix is in this PR https://github.com/mlcommons/algorithmic-efficiency/pull/810, which I'll merge in separately before mergin this in.

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/randaugment.py
```diff
@@ -8,7 +8,8 @@
 import math
 
 import tensorflow as tf
-from tensorflow_addons import image as contrib_image
+
+#from tensorflow_addons import image as contrib_image
```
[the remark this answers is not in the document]
init-22: Hello @runame , this is because tensorflow_addons is not supported now, see this : https://www.tensorflow.org/addons/tutorials/image_ops

Also, the current TF version is 2.18 but this requires < 2.15, we might have to re-write the functions used in the current code for rotate, translate, shear operations or use TF 2.15

algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py
init-22: Yeah so the function `corpus_bleu` which we use for calculating the score is taken from belu.py and belu.py is dependent on sacrebleu's 1.3.1 version : https://github.com/mjpost/sacrebleu/blob/v1.3.1/sacrebleu.py
Current verision of the package is at 2.4.3. 
Current version (1.3.1) and belu.py gives errors because of this; they've changed the location of some of the variables and modified the __init__.py

In either case we need to install sacrebelu, because belu.py still require to use sacrebelu 1.3.1 so why not just use the latest methods and call the function from there so instead of bleu_score = bleu.corpus_bleu(predictions, [references]).score

I am currently doing:
bleu_score = sacrebleu.corpus_bleu(predictions, [references]).score

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/randaugment.py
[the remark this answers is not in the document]
runame: Thanks for the explanation. Do I understand correctly that the code is not runnable then (on this branch)?

algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py
runame: The bleu score calculation in the current version of `sacrebleu` is not equivalent to what we used before, hence our manual solution (see #264). So if we want to ensure that bleu scores are comparable across different versions of the benchmark, `sacrebleu` is an exception and should not be updated, i.e. it should stay pinned to 1.3.1.

priyakasimbeg: @init-22 could you check that pinning it to 1.3.1 does not break anything and revert the change to the bleu_score calls?

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/randaugment.py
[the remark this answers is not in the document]
init-22: Yes, the code wont run incase you want to use the functions from tensorflow_addons, I think the CI checks didnt require it so it passed the checks and other tests, we have to write a custom logic or use another package for the same, shall we stick to TF2.15 then? but then the numpy will also change I think it needs numpy <2

algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py
init-22: yeah makes sense, will get back to you on this one

init-22: @runame @priyakasimbeg 
Changed it back to 1.3.1, all checks are also passed.

Also, I remove the sacrebleu package dependency altogether in my local machine, I kept only the relevant part from sacrebleu v1.3.1 and ran the code and it worked too! keeping all global vars and functions as the original one, this way we dont need to install sacrebleu!

If you think, that's something we should do, I can raise another PR specific to that change once this branch is merged.

init-22:
@priyakasimbeg Please check the output of pip freeze

[pip_freeze_python311.txt](https://github.com/user-attachments/files/18049140/pip_freeze_python311.txt)

algorithmic_efficiency/workloads/wmt/wmt_jax/workload.py
priyakasimbeg: Ok sounds good!

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/randaugment.py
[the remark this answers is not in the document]
runame: Maybe we should upgrade TF in a separate PR that also adjusts the code to work and be functionally equivalent to the previous one?

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/randaugment.py
```diff
@@ -8,7 +8,10 @@
 import math
 
 import tensorflow as tf
-from tensorflow_addons import image as contrib_image
+
+from .custom_tf_addons import rotate_img
+from .custom_tf_addons import transform
+from .custom_tf_addons import translate
```
runame, line 14: Can you use absolute paths here?

runame (approved):
LGTM, thanks for porting the custom tf addons!

Maybe we can get rid of the `sacrebleu` dependency in a follow up PR.

algorithmic_efficiency/workloads/imagenet_resnet/imagenet_jax/randaugment.py
init-22: Sure! done.

runame: Thanks!

PR #823, opened 2024-12-19
Fsdp pytorch draft PR

davidtweedle:
A draft PR for potential changes for the pytorch workloads to upgrade to FSDP (fully sharded data parallel) from DDP (distributed data parallel).

Summary for changes to: cifar, mnist, criteo1tb, imagenet vit, imagenet resnet, librispeech deepspeech, librispeech conformer, ogbg, wmt, fastmri
- import required packages (e.g., fsdp)
- construct model using FSDP constructor instead of DDP constructor
- very naive sharding for now
- Crucially: must zero grad before eval 

Summary for changes to momentum (as simple test optimizer):
- first compute weighted loss on each device
- then loss.backward (the gradient of the losses will now be all reduced by a pytorch communication hook)
- then display the correct loss

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ __pycache__
 .vscode/
 env/
 venv/
+.venv/
 workdir/
 makefile
 *.out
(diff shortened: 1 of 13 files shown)


priyakasimbeg:
won't fix. closing for now

PR #828, opened 2024-12-22
Removed the sacrebleu dependency

init-22:
This PR removes the dependency on the sacrebleu package from the codebase. The only required functionality, corpus_bleu, has been directly incorporated by referring to the relevant implementation from sacrebleu version 1.3.1.

Reference:
https://github.com/mjpost/sacrebleu/blob/v1.3.1/sacrebleu.py


priyakasimbeg:
Great! thanks

Issue #803, opened 2024-10-31
Improving package setup to pyproject.toml


IFFranciscoME:
## Description

*Improve the approach for packaging and use a `pyproject.toml` approach*

## Context
The common approach for a python project to be used is either a direct way, mostly like cloning the repository, and a compacted way, mostly in the form of packaging and/or containerization. In either case, for the same version of the software, the same functionality should be available for any given system independently on how it was installed. 

## Problem
When the user decides to go for the package/container route of installing the software, there will be dependency issues not straight forward solvable for some cases (more yet to be mapped). A non exhaustive list of these problems is:

- Python version and sub-version compatibility.
- `PYTHONPATH` values maybe are not updated programatically.
- Order of dependency installs might break out inter-dependencies.
- Dependencies to download data might conflict with those for running workloads.

## Elements for the solution (Draft)

In general, it might me a good opportunity to update the packaging from a `setup.py` oriented approach to `pyproject.toml` approach. 

[PEP 518 – Specifying Minimum Build System Requirements for Python Projects](https://peps.python.org/pep-0518/)

> This PEP specifies how Python software packages should specify what build dependencies they have in order to execute their chosen build system. As part of this specification, a new configuration file is introduced for software packages to use to specify their build dependencies (with the expectation that the same configuration file will be used for future configuration details).

[Externally Managed Environments](https://packaging.python.org/en/latest/specifications/externally-managed-environments/#externally-managed-environments)

> To allows a Python installation to indicate to Python-specific tools such as pip that they neither install nor remove packages into the interpreter’s default installation environment

Some other details might be useful:

- Every package now should be strictly necessary (at least really really hard to not depend on it)
- Use of exact versions for every single package.
- Include explicit installation order and without using cache installed versions.
- Make sure the latest version of pip is installed.
- Use of installation flags for python packages.
- Compile/store the package's Wheel (offline file) according to particular architectures.

IFFranciscoME:
Ok, @priyakasimbeg, here is my proposition of more actionable items to start the first phase of refactoring: And is in the Datasets installation/downloading process actually, previous to the overall project.

Problems:

- Dependencies across datasets might differ in time.
- Dependencies across datasets and the submission script, or any other script for that matter, might differ in time.
- Pre-processing for one dataset might be different to another one.
- Changes and updates into the understanding for each dataset might be at different pace.

Improvement oportunities: 

1. Move from a monolithic dataset config to a per-data set config.
[...]

PR #830, opened 2025-01-15
Switch to `pyproject.toml`

fsschneider:
This PR replaces the current `setup.cfg` and `setup.py` with (hopefully) equivalent `pyproject.toml`.
As described in #803 this is the PEP standard for package setup.
It probably doesn't fully resolve said issue as it also included more proposed changes du the dataset setup.

Note: We plan to move to `ruff` (replacing `isort`, `pylint` and `yapf`) as well as renaming the package to `algoperf`. Therefore, the `pyproject.toml` will change substantially with follow-up PRs.

@IFFranciscoME, I currently can't add you as a reviewer, but I am happy to hear your feedback as well.


IFFranciscoME:
@fsschneider Thanks for the tag and the acknowledgement of the 2 cents of work I've previously added to this engineering topic. 

From a quick look, here are some notes

1. 100% agree in going for `pyproject.toml`, which I think is a good starter for a folder restructure as well. 

2. Great to see the `[project.optional-dependencies]`

3. Perhaps "compiled examples" could be a good thing, i.e. to have an already compiled version of the package, for a "standard" system architecture-config. 

I will try to make more free space for this, have had a lot of focus on some PhD work in the past 2 months, but hey, time in one's agenda is not to be founded but has to be created.

priyakasimbeg:
Just checking, this doesn't require any updates to our documentation (instructions for installation) right?

fsschneider:
No, I don't think so. It is still pip installable as before and all commands are identical.

PR #832, opened 2025-01-15
Minor nits

fsschneider:
Depends on #830 and #831

- Rename a job in `regression_tests_variants`. There were two jobs called `criteo_resnet_pytorch` I believe the latter one should actually be called `criteo_embed_init_pytorch`.
- Minor nits to formatting in markdown files (to be consistent with what `markdownlint` suggests).


priyakasimbeg:
The regression test for variants are in an incomplete state. I think we should probably add all the other workloads to the test or remove the entire test.
Since we will not be using the variants for scoring, we may not need to maintain (and test them). Although if we don't maintain them, we should probably just remove them. Wdyt Frank?

fsschneider:
Yeah, I agree. We probably don't want to maintain them and should thus remove them from the code.
We could double-check this decision in a quick working group decision. And then in a follow-up PR, we could remove the held-out workloads from our codebase.

PR #833, opened 2025-01-15
Rename package to `algoperf`

fsschneider:
Depends on #830, #831, and #832.

This PR renames the package from `algorithmic_efficiency` to simply `algoperf`.

Note that the _repository_ is still called `algorithmic-efficiency` (with a hyphen instead of an underscore). We could consider renaming the repo as well (to `algoperf`, `algoperf_codebase`, or `algoperf_benchmark`) but this is under MLCommons' control so we would need to contact them.


fsschneider:
@priyakasimbeg I accidentally created this PR to merge with `main` which triggered the containerized regression and training difference tests. I have canceled these tests and switched to merge with `dev`. But the canceled tests still show up as `failed`, however all non-canceled tests are passing.

PR #835, opened 2025-01-18
fix: fixing the range of np.int32

init-22:
#801


init-22:
Closing this because it is already resolved

PR #844, opened 2025-02-11
fix: frozen dict conversion

init-22:
tests/test_traindiffs.py was failing, this PR aims at solving the error in that by changing the normal python dictionary to FrozenDict of jax and and using python instead of python3


priyakasimbeg (approved):
Awesome, thank you for this fix Isaac.

PR #848, opened 2025-03-06
Migrate JAX workloads from pmap to jit

priyakasimbeg:
## Purpose

The goal of this PR is to allow model parameter and optimizer state sharding, and also to migrate the JAX code from using jax.pmap to using jax.jit. 

## TODOs:
- [x] Migrate reference optimizers to use jax.jit
  - [X] Nesterov
  - [X] AdamW
  - [x] Others
- [x] Migrate workloads to use jax.jit
  - [X]  (Test workload) MNIST
  - [X]  (Test workload) CIFAR
  - [x] WMT
  - [x] Criteo1TB
  - [x] FastMRI
  - [x] Librispeech
  - [x] OGBG
  - [x] ImageNet
 
## Changelog
- Added some sharding utilities to handle data distributed
- Replaced pmap code for all workloads with jit
- Modified reference submissions accordingly
- Updated checkpoint and data_utils to support the new approach (mostly removing explicit jax_utils.replicate calls).
- Upgraded Jax version to 0.7.0
- Factored out the sharding into shape (bsz/n_devices, ...) out of `shard_and_maybe_pad`. PyTorch workloads now call a separate shard function if they used `shard_and_maybe_pad` before

### DIFF
--- a/tests/reference_algorithm_tests.py
+++ b/tests/reference_algorithm_tests.py
@@ -307,7 +307,7 @@ def _make_one_batch_workload(
 
         if framework == 'pytorch':
           fake_batch_iter = ogbg_input_pipeline._get_batch_iterator(
-            _fake_iter(), global_batch_size, shard=True
+            _fake_iter(), global_batch_size, shard_dim=True
           )
         else:
           fake_batch_iter = ogbg_input_pipeline._get_batch_iterator(


algoperf/workloads/fastmri/fastmri_jax/workload.py
```diff
@@ -154,13 +156,12 @@ def _eval_model_on_split(self,
           num_batches=num_batches)
 
     total_metrics = {'ssim': 0., 'loss': 0.}
```
priyakasimbeg, line 158: why did we swap out the eval_rngs with the model rng?

rka97:
We didn't migrate all the optimizers. Should we look into migrating the following?

- adafactor
- lamb
- sam
- shampoo

PR #850, opened 2025-03-12
Fix `prepare_for_eval` usage in self-tuning track

Niccolo-Ajroldi:
I noticed that `prepare_for_eval` was not passed as an argument to `train_once` in the self tuning track, sorry for the oversight.

It's a 1-line fix.


priyakasimbeg:
thanks for catching this!

PR #851, opened 2025-03-12
Allow self-tuning submissions to specify `dropout_rate` and `aux_dropout_rate`

### LINKED ISSUE #753: Models will always be initialized without dropout layers in self-tuning ruleset
In submission_runner.py, if we are in the self-tuning rules, the hyperparameters argument to train_once will always be None.

Then in this code snippet

```
    dropout_rate = None
    aux_dropout_rate = None
    if hasattr(hyperparameters, 'dropout_rate'):
      dropout_rate = hyperparameters.dropout_rate
    if hasattr(hyperparameters, 'aux_dropout_rate'):
      aux_dropout_rate = hyperparameters.aux_dropout_rate
    model_params, model_state = workload.init_model_fn(
        model_init_rng, dropout_rate, aux_dropout_rate)
```

workload.init_model_fn will always get None for dropout_rate and aux_dropout_rate, so Dropout layers won't ever be added to the model.

Although submissions could call workload.init_model_fn again themselves to make use of its side effect of setting workload._model, this is awkward and also challenging for workloads near the memory limit since it involves superfluously reconstructing model_params again on device.

Niccolo-Ajroldi:
## Original issue
Self-tuning submissions use default values for `dropout_rate` and `aux_dropout_rate` (see #753).
We want to allow them to specify custom values for these hyperparameters.

## Solution
Allow self-tuning submission to provide a JSON file specifying values for these hyperparameters. 
The JSON file should look like this:
```json
{
    "dropout_rate": 0.1,
    "aux_dropout_rate": 0.2,
}
```

This is passed to `submission_runner.py` through `--tuning_search_space`. 

We read from such JSON file, and check that:
1. it specifies a value only for the _allowed hyperparameters_, and that 
2. it specifies _one single value_ for each of them. 

We then proceed to call `train_once` with the correspondent fixed hyperparameters.

## Discussion
I think this solution is pretty clean, I am just dubious about passing the JSON file path trough `tuning_search_space`., as this might create some confusion. Perhaps we shall change the name of this flag, or we can add a different flag?

Fix #753


priyakasimbeg:
I agree with you, I think we should add an additional flag that is optional for self-tuning and disallowed for external-tuning ruleset (so raise an error if the value of the flag is not none for external-tuning). 
Perhaps we can call this flag self_tuning_init_hyperparemeters_config? Maybe that is a bit long, do you have any suggestions for names?

priyakasimbeg:
Underlying issues addressed in of https://github.com/mlcommons/algorithmic-efficiency/pull/875

PR #852, opened 2025-03-13
Fix self-tuning submissions hyperparameter specification

Niccolo-Ajroldi:
Self-tuning submissions in `prize_qualification_baselines/self_tuning/` define `HPARAMS` as a dictionary:
```python
HPARAMS = {
    "dropout_rate": 0.1,
    "learning_rate": 0.0017,
    ...
}
```
But later try to access its entries via attribute access:

https://github.com/mlcommons/algorithmic-efficiency/blob/5cc62ea286403add41b0e7929bf638681b850718/prize_qualification_baselines/self_tuning/pytorch_nadamw_full_budget.py#L211-L217

This pull request just converts `HPARAMS` to a named tuple to allow for attribute access.

### DIFF
--- a/prize_qualification_baselines/self_tuning/jax_nadamw_full_budget.py
+++ b/prize_qualification_baselines/self_tuning/jax_nadamw_full_budget.py
@@ -1,5 +1,6 @@
 """Submission file for an NAdamW optimizer with warmup+cosine LR in Jax."""
 
+import collections
 import functools
 
 # isort: off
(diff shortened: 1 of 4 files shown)


priyakasimbeg (changes requested):
Thanks for catching this! It looks like in the jax_* files the hyperparemters are indexed as dicts already. So let's just convert the hyperparemeters in the the pytorch_* qualification baselines to namedtuples.

Niccolo-Ajroldi:
@priyakasimbeg thanks for the comment! I fixed the jax ones. Should be ready to merge now (:

PR #856, opened 2025-03-19
LM workload - first steps

Niccolo-Ajroldi:
## Language Modeling Workload - First Steps

Changes:
- Implemented LM workload base class definition, and JAX and PyTorch subclasses
- Renamed `datasets` into `dataset` to avoid conflict with huggingface [datasets](https://huggingface.co/docs/datasets/en/index) library. Provisionary name, we'd need to update the docs accordingly.
- Added `download_finewebedu` to `dataset/dataset_setup.py`: this function downloads, tokenize and chunks data in blocks of seq_len+1. It uses [batch mapping](https://huggingface.co/docs/datasets/en/about_map_batch) and multiprocessing to speed up tokenizaation and concatenation/chunking.
- Implemented dataloaders in JAX and PyTorch


priyakasimbeg:
Thanks Niccolo! 
Could just you run the yapf and isort tools to fix the failing tests?

Niccolo-Ajroldi:
Fixed!

PR #857, opened 2025-03-19
Jit switch for ogbg

davidtweedle:
I added some code to the ogbg workload according to the migration guide helpfully provided by Ahmed. Unfortunately, there seems to be some bugs that I still hope to fix.


davidtweedle:
Here is a log from attempting to run the jitted ogbg workload:
[ogbg.log](https://github.com/user-attachments/files/19349644/ogbg.log)

PR #860, opened 2025-04-03
[WIP] LM Workload

rka97:
This is for the LM workload.


fsschneider:
I might be wrong here, but I just randomly saw that we are using a default MLP expansion factor of $4$ (i.e., https://github.com/mlcommons/algorithmic-efficiency/blob/5c85c7e278ffa540d65b1d49f0bd1d0cad732052/algoperf/workloads/lm/lm_pytorch/workload.py#L40).

However, it seems to me, that we are also using a gated linear unit, which uses an initial expansion that is twice as large (since it is split into the gate and the value), i.e. https://github.com/mlcommons/algorithmic-efficiency/blob/5c85c7e278ffa540d65b1d49f0bd1d0cad732052/algoperf/workloads/lm/lm_pytorch/plainlm_model.py#L28-L30

I believe that the common expansion factor of $4$ is for models that use non-gated activation functions (e.g., original Transformer with ReLU or GPT-2 with GELU). To keep the number of parameters constant, people commonly divide by $2/3$ for gated versions. This keeps the total number of parameters in the MLP the same (some small approximation error due to the non-integer multiplication).

If you agree, we should either change the default to $4/3$, or adjust the computation of the `hidden_dim` in the MLP to account for this scaling.

@rka97 @priyakasimbeg @Niccolo-Ajroldi

Niccolo-Ajroldi:
@fsschneider agree, we should definitely adjust the expansion factor as you suggested. This was probably an oversight, as in plainLM we rescale when using GLU.

priyakasimbeg:
After discussing with Frank, we've decided to handle the modification to the dense layer dimensions in the MLP block itself that adjusts the dimensions based whether we use the gated unit.

I will update the jax workload. This workload is currently ahead of the pytorch one on PR 891. @Niccolo-Ajroldi  I will add you to that PR so you get a good sense of what needs to be updated.

algoperf/workloads/finewebedu_lm/input_pipeline.py
```diff
@@ -0,0 +1,148 @@
...
+    ds = batch_with_padding(
+      sequences_ds,
+      batch_size,
+      padded_shapes={
+        'inputs': (batch_size, None),
+        'targets': (batch_size, None),
+      },
+    )
+    ds = ds.take(num_batches) if num_batches is not None else ds
+    ds = ds.repeat()
```
priyakasimbeg, line 144: @rka97 I don't think we want a repeat here, unless there is something else I'm missing. For the evaluations (both train_eval and validation splits) we should be careful not to repeat evals on examples. For example if we evaluate on the validation set, that means we should account for the predictions on the example in the validation set exactly once. 
If we try to  take more examples than we have available in the split we should raise a StopIteration error, which will happen naturally when the ds runs out of examples.

algoperf/workloads/finewebedu_lm/input_pipeline.py
```diff
@@ -0,0 +1,148 @@
...
```
priyakasimbeg, line 125: Same comment as below. I don't think we want to repeat the examples in the eval_train train split.

priyakasimbeg (changes requested):

algoperf/workloads/finewebedu_lm/input_pipeline.py
rka97: so in eval_model_on_split, we do pass the number of batches

```python

  def _eval_model_on_split(
    self,
    split: str,
    num_examples: int,
    global_batch_size: int,
    params: spec.ParameterContainer,
    model_state: spec.ModelAuxiliaryState,
    rng: spec.RandomState,
    data_dir: str,
    global_step: int = 0,
  ) -> Dict[str, float]:
    """Run a full evaluation of the model."""
    num_batches = int(math.ceil(num_examples / global_batch_size))

    # Handle edge case where num_batches is 0 (e.g., test split with 0 examples)
    if num_batches == 0:
      return {'loss': 0.0, 'ppl': 1.0}

    if split not in self._eval_iters:
      # These iterators will repeat indefinitely.
      self._eval_iters[split] = self._build_input_queue(
        rng, split, data_dir, global_batch_size, num_batches=num_batches
      )
``` 

The above is called by workload.eval_models:
```python

  def eval_model(
    self,
    global_batch_size: int,
    params: ParameterContainer,
    model_state: ModelAuxiliaryState,
    rng: RandomState,
    data_dir: str,
    imagenet_v2_data_dir: Optional[str],
    global_step: int,
  ) -> Dict[str, float]:
    """Run a full evaluation of the model."""
    logging.info('Evaluating on the training split.')
    train_metrics = self._eval_model_on_split(
      split='eval_train',
      num_examples=self.num_eval_train_examples,
      global_batch_size=global_batch_size,
      params=params,
      model_state=model_state,
      rng=rng,
      data_dir=data_dir,
      global_step=global_step,
    )
    eval_metrics = {'train/' + k: v for k, v in train_metrics.items()}
    # We always require a validation set.
    logging.info('Evaluating on the validation split.')
    validation_metrics = self._eval_model_on_split(
      'validation',
      num_examples=self.num_validation_examples,
      global_batch_size=global_batch_size,
      params=params,
      model_state=model_state,
      rng=rng,
      data_dir=data_dir,
      global_step=global_step,
    )
    for k, v in validation_metrics.items():
      eval_metrics['validation/' + k] = v
    eval_metrics['validation/num_examples'] = self.num_validation_examples
    # Evaluate on the test set. TODO(znado): always eval on the test set.
    try:
      if self.num_test_examples is not None:
        logging.info('Evaluating on the test split.')
        test_metrics = self._eval_model_on_split(
          'test',
          num_examples=self.num_test_examples,
          global_batch_size=global_batch_size,
          params=params,
          model_state=model_state,
          rng=rng,
          data_dir=imagenet_v2_data_dir if imagenet_v2_data_dir else data_dir,
          global_step=global_step,
        )
        for k, v in test_metrics.items():
          eval_metrics['test/' + k] = v
        eval_metrics['test/num_examples'] = self.num_test_examples
    except NotImplementedError:
      pass

``` 
which passes the number of validation examples to the workload. This way: we repeat each example exactly once, since the for loop will terminate after num_batches iterations. This allows us to not have to reset the eval dataset loader every time. This seems to be the approach used in other workloads, e.g. in ImageNet [dataloader](https://github.com/mlcommons/algorithmic-efficiency/blob/91988af436f452021e98a61e5144a89d14418e20/algoperf/workloads/imagenet_resnet/imagenet_pytorch/workload.py#L149)
```python
    dataloader = torch.utils.data.DataLoader(
      dataset,
      batch_size=ds_iter_batch_size,
      shuffle=not USE_PYTORCH_DDP and is_train,
      sampler=sampler,
      num_workers=4 if is_train else self.eval_num_workers,
      pin_memory=True,
      drop_last=is_train,
      persistent_workers=is_train,
    )
    dataloader = data_utils.PrefetchedWrapper(dataloader, DEVICE)
    dataloader = data_utils.cycle(
      dataloader,
      custom_sampler=USE_PYTORCH_DDP,
      use_mixup=use_mixup,
      mixup_alpha=0.2,
    )

    return dataloader
```

rka97:
Adding some TODOs:

- [X] Add model matching tests (i.e. for the same inputs, outputs should be the same, kinda like [model_match](https://github.com/rka97/model_match))
- [X] Fix initialization to be the same for both PyTorch and JAX (there are some minor differences it seems)
- [ ] Add integration test for the lm workload to github actions

algoperf/workloads/finewebedu_lm/finewebedu_lm_jax/models.py
```diff
@@ -0,0 +1,397 @@
+"""
+Originally based on code from the NanoDO repository under the Apache 2.0 license:
```
priyakasimbeg, line 2: Can we rename this file to models.py to be consistent with the pattern in the other workload definitions.

algoperf/workloads/finewebedu_lm/finewebedu_lm_pytorch/models.py
```diff
@@ -0,0 +1,344 @@
+"""
+Originally based on the plainLM codebase:
```
priyakasimbeg, line 2: Can we rename this file to models.py to be consistent with the other workload definitions

algoperf/workloads/workloads.py
```diff
@@ -113,6 +113,7 @@
     'workload_path': 'librispeech_deepspeech/librispeech',
     'workload_class_name': 'LibriSpeechDeepSpeechNormAndSpecAugWorkload',
   },
+  'lm': {'workload_path': 'lm/lm', 'workload_class_name': 'LmWorkload'},
```
priyakasimbeg: Now that we have all the important implementation details figured out should we pick a more descriptive name for the workload? I am thinking perhaps 'fineweb_edu_lm'?

algorithms/archived_paper_baselines/adamw/pytorch/submission.py
```diff
@@ -189,6 +189,8 @@ def get_batch_size(workload_name):
     return 128
   elif workload_name == 'mnist':
     return 16
+  elif workload_name == 'lm':
+    return 4
```
priyakasimbeg: This should work for bsz 64 right?

algorithms/archived_paper_baselines/nesterov/jax/submission.py
```diff
@@ -292,6 +292,8 @@ def get_batch_size(workload_name):
     return 16
   elif workload_name == 'cifar':
     return 128
+  elif workload_name == 'lm':
+    return 8
```
priyakasimbeg: Should work for bsz 64 right?

priyakasimbeg (changes requested):
Second round of small requested changes.

Perhaps something we should discuss, we need a more descriptive name for the workload. E.g. fineweb_edu_lm. What do you all think? @Niccolo-Ajroldi  @rka97

algoperf/workloads/finewebedu_lm/input_pipeline.py
priyakasimbeg: OK you are correct. I didn't realize we passed the number of examples when we construct the splits as well. Thanks for your thoroughness!

algoperf/workloads/finewebedu_lm/finewebedu_lm_jax/models.py
Niccolo-Ajroldi: Agree.

algoperf/workloads/finewebedu_lm/finewebedu_lm_pytorch/models.py
Niccolo-Ajroldi: Agree.

algoperf/workloads/workloads.py
Niccolo-Ajroldi: `fineweb_edu_lm` or `finewebedu_lm` make sense, matching the other workload names.

algorithms/archived_paper_baselines/nesterov/jax/submission.py
rka97: Yes, I just made it smaller because I was debugging with v100s. Changed it back to 64.

algorithms/archived_paper_baselines/adamw/pytorch/submission.py
rka97: Yes, I just made it smaller because I was debugging with v100s. Changed it back to 64.

algoperf/workloads/workloads.py
rka97: Changed to finewebedu_lm.

algoperf/workloads/finewebedu_lm/finewebedu_lm_jax/models.py
rka97: Done.

algoperf/workloads/finewebedu_lm/finewebedu_lm_pytorch/models.py
rka97: Done.

Niccolo-Ajroldi (dismissed):

PR #871, opened 2025-06-05
Update metrics.py - fix for ogbg pytorch

davidtweedle:
It seems that the problem affecting the pytorch ogbg workloads (but only if they run for some length of time) has to do with jax/xla cpu compilation of the metrics computation. By converting the jax arrays to numpy, hopefully this can be avoided. The next step is to test on schedule free and shampoo, which I hope to do very soon.


davidtweedle:
You can see here that schedule free now completes the run with the changes.
https://pastebin.com/RgqMqgkb

davidtweedle:
Tested again but this time only replacing the call to jax for sigmoid with numpy sigmoid. This also avoids the crash. So the problem seems to be calling jax sigmoid in pytorch.

PR #882, opened 2025-08-14
Schedule free AdamW 

init-22:
This PR has Schedule free adamw implementation both in Pytorch and Jax


priyakasimbeg:
As discussed offline, let's move this PR to the submissions repository.

PR #891, opened 2025-10-06
Updates to LM PR

priyakasimbeg:


algoperf/workloads/lm/lm_jax/workload.py
```diff
@@ -101,41 +82,58 @@ def model_fn(
...
+    Returns:
+      {'summed': scalar summed loss, 'n_valid_examples': scalar number of
+      valid examples in batch, 'per_example': 1-d array of per-example losses}
+    """
+    if logits.ndim != targets.ndim + 1:
+      raise ValueError(
+        f'Incorrect shapes. Got shape {logits.shape} logits and '
+        f'{targets.shape} targets.'
+      )
+    smoothed_targets = optax.smooth_labels(
```
rka97: Will calling onehot add a lot of extra memory? Since targets are already ids, we can instead use the old way without adding extra memory.

algoperf/workloads/lm/lm_jax/workload.py
```diff
@@ -145,8 +143,10 @@ def _eval_batch(self,
     """Evaluate the model on a single batch."""
     logits, _ = self.model_fn(
         params, batch, model_state, spec.ForwardPassMode.EVAL, rng, False)
-    targets = batch['targets']
-
     # Calculate cross-entropy loss
-    loss = -jnp.sum(targets * jax.nn.log_softmax(logits, axis=-1))
-    return loss
+    # TODO(kasimbeg): add weights?
+    metrics = self.compute_weighted_cross_entropy(logits, batch['targets'], batch['weights'])
```
rka97: For this workload I don't think we need weights for the cross-entropy calculation. Maybe we should explicitly del any weights?

algoperf/workloads/lm/input_pipeline.py
```diff
@@ -1,154 +1,138 @@
...
+  ds = get_lm_dataset(data_rng, split, data_dir, global_batch_size, num_batches)
+  
+  it = map(
+    functools.partial(
+      data_utils.shard_and_maybe_pad_np, global_batch_size=global_batch_size
+    ),
+    ds,
+  )
+
+  return iter(it)
```
rka97, line 69: Should we use itertools.cycle here?

algoperf/workloads/lm/lm_pytorch/workload.py
rka97, line 136: We should remove _eval_batch in this file. The workload.py version already does the same thing.

algoperf/workloads/lm/lm_jax/workload.py
```diff
@@ -101,41 +82,58 @@ def model_fn(
...
+  def _normalize_eval_metrics(
+    self, num_examples: int, total_metrics: Dict[str, Any]
+  ) -> Dict[str, float]:
+    """Normalize eval metrics."""
+    del num_examples
+    eval_denominator = total_metrics.pop('denominator')
+    return jax.tree.map(lambda x: float(x / eval_denominator), total_metrics)
+
 
   def _eval_batch(self,
```
rka97: The _eval_batch in the common workload.py should work for here, do we need this function?

algoperf/workloads/lm/lm_jax/workload.py
rka97: Added a commit below to fix this. Maybe we should also use it for WMT? I was getting OOM errors on the V100s without it.

priyakasimbeg: ok sounds good, thanks. One thing I forgot to mention is that we want to support label smoothing which is another reason I refactored the loss function calls.

algoperf/workloads/lm/lm_jax/workload.py
priyakasimbeg: The weights are used to identify padded elements in the validation split and correctly calculate the number of tokens returned in the eval dict.

algoperf/workloads/lm/input_pipeline.py
priyakasimbeg: No I don't think so. The input_pipeline already calls .repeat() on the train split and we don't want cycle on the validation split.

algoperf/workloads/lm/lm_jax/workload.py
priyakasimbeg: yes you're right, the shared workload version did not return the expected eval_dict. I replaced the shared version with this version now and removed the _eval_batch from the jax-specific workload definition.

algoperf/workloads/lm/lm_pytorch/workload.py
priyakasimbeg: Left a todo in the pytorch workload to clean this up

PR #897, opened 2026-01-12
Fix slow ImageNet workloads on PyTorch

rka97:
1. Change the number of workers to be larger.
2. Add caching so that we don't have to rescan the file system every time we run the PyTorch workload (made a big difference while debugging!).
3. Use torch built-in attention for the vision transformer. The custom module was a lot slower than JAX's built-in attention on the A100 (but crucially not on the V100, where they were the same speed!).
4. Compile the ViT workload as well, without compilation it's a lot slower than JAX.
5. Add a test that measures the speed difference between the jax and pytorch workloads. Currently this depends on conda envs, but we can modify it to use docker instead.


submission_runner.py
```diff
@@ -409,6 +409,21 @@ def train_once(
       train_state['training_complete'] = True
 
     train_step_end_time = get_time()
+    if global_step == 11:
```
priyakasimbeg: This looks like it was more for debugging right?

PR #900, opened 2026-01-27
LM Workload TF32

rka97:
This is the PR for the LM workload with mixed-precision training on 4xA100.


priyakasimbeg:
This branch has the mixed precision code changes for all of the workloads. We have an existing PR open from the lm_workload_base branch that only has the LM workload changes. Note I did not do a final test with the mixed precision for just the LM workload after increasing the number of evals to get a better timing estimates. At this point we may just opt to do LM workload with TF32 for the new release since we don't have anymore bandwidth to test changes.

PR #915, opened 2026-03-09
Minor fixes

aahladc:
Don't merge. 

This is for testing purposes only.
Scoring changes.


run_pytorch_scoring.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
priyakasimbeg, line 1: Can you move this file into utils/slurm

pytorch_scoring_config_1.json
```diff
@@ -0,0 +1,299 @@
+{
```
priyakasimbeg, line 1: I don't think we want to commit this file to the repository. Can you remove it from this branch?

scoring/utils/slurm/run_jobs.sh
```diff
@@ -43,41 +33,137 @@ then
...
+      exit 1
+      ;;
+  esac
+done
+
+# Pull docker image
+yes | gcloud auth configure-docker "$REPO"
+docker pull "$IMAGE"
+
+# Set variables from config file if not already set by flags
```
priyakasimbeg, line 125: So this script will now take flags for all of the config fields, that will override the config fields. Why do we need this?

priyakasimbeg (changes requested):
Could you update the description of the PR? It currently still says 'do not merge'. 

Left some comments
-  I don't think should  to check in any json file outputs from the make_job_config.py
- the sbatch scripts should all be under utils/slurm
- I'm a bit confused why we need flags for all the config options in the run_jobs sbatch script. The nice thing about having the config is that we have a complete and reproducible record of the options we passed in.

